PR #5, opened 2026-06-23
docs(skill): backend test authoring rules (call the test fn + sandbox limits)

fyZhang66:
## Why
Debugging real CLI backend tests surfaced two authoring gaps that produced misleading results:

1. **Defined-but-uncalled tests silently passed.** The runner executes `--code-file` top-to-bottom (not `pytest`), so a `def test_*(): assert …` that's never called never runs its assertions → green regardless.
2. **Imports outside the sandbox fail to run.** The execution environment only has stdlib + `requests` + `pytest` + `numpy` + `scipy`. Tests that `import` the project's own source modules or other packages (`torch`, `pandas`, local app modules) die with `ModuleNotFoundError`.

## Change (skill assets only — markdown)
- `skills/testsprite-verify.skill.md`: backend example now ends with a module-level call to the test fn, plus a new **"Execution environment (backend)"** note (supported package set + "test the API over HTTP with `requests`, don't import app source / unavailable packages").
- `skills/testsprite-verify.codex.md`: two terse mirroring bullets.

No code/behavior change.

## Related
The execution image is gaining `numpy`/`scipy` so the "supported set" above is accurate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skills/testsprite-verify.codex.md
+++ b/skills/testsprite-verify.codex.md
@@ -64,6 +64,8 @@ testsprite test run --all --project <id> [--filter <substr>] \
   already have the change deployed (e.g. a CI preview deploy) — the CLI tests a
   deployed URL, it doesn't host your environment. Running earlier verifies the
   previous build.
+- Backend `--code-file`: the runner executes the file top-to-bottom (not `pytest`), so **call your `test_*` function(s) at the end of the file** — a defined-but-uncalled test silently passes.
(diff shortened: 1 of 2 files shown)


fyZhang66:
Closing — this backend-test authoring guidance belongs in the internal testsprite-cli dev lineage (src/assets/agent-skill/gundam.*), not the public repo's skills/.

PR #6, opened 2026-06-23
fix(test): prevent batch-run scheduler from serializing after first wave

crypticsaiyan:
## Summary
 
The `create-batch --run` scheduler (`runBatchRun` in `src/commands/test.ts`) correctly launches the initial `concurrencyLimit` jobs in parallel. However, the steady-state loop inadvertently awaits each subsequent job's full completion (both the trigger and the `--wait` poll) before launching the next one. 

As a result, the effective concurrency drops to `1` after the first wave completes, ignoring the `--max-concurrency` flag for the remainder of the run.

**Note:** 3 of the 4 fan-outs in this file already use the correct launch-then-relaunch pattern; this specific scheduler was the outlier.

## The Fix
Replaced the blocking `drainOne()` semaphore logic with a non-blocking `startNext()` callback pattern. 

This removes the blocking `await` that was stalling the `Promise.race` queue. It mirrors the implementation used by its sibling `pollFreshAccepted`, ensuring we launch up to the concurrency limit, immediately relaunch a new job into a slot as soon as one completes, and never block the scheduler on a single job's completion.

## Testing & Validation

 **Regression Test:** Added a new test case in `src/commands/test.test.ts` using equal-delay mock responses. 
  - **Before fix:** Fails (active job count drops to 1 for tail jobs).
  - **After fix:** Passes (active job count correctly remains at 3).
 
  - `npm run lint` & `npm run format:check`
  - `npm run typecheck`
  - `npm test` (1448 tests passing)
  - `npm run test:coverage` (≥80% threshold met)

**Files Changed:**
- `src/commands/test.ts`
- `src/commands/test.test.ts`

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -6694,6 +6694,103 @@ describe('runCreateBatch', () => {
     ).resolves.toBeDefined();
   });
 
+  // Regression test: create-batch --run must keep launching new triggers
+  // up to --max-concurrency as slots free up, not collapse to serial
+  // after the first wave. Uses equal-delay trigger responses so the
+  // first three jobs settle in the same microtask batch, the exact
+  // condition that exposed the bug (race only reacts to one settlement,
+  // then blocks the scheduler on the whole next job before moving on).
+  it('--run keeps concurrency at --max-concurrency for tail jobs, not just the first wave', async () => {
+    const { credentialsPath } = makeCreds();
+    const specs = Array.from({ length: 6 }, (_, i) => ({ ...FE_SPEC, name: `spec-${i}` }));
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Verified locally — the new startNext() relaunch pattern keeps concurrency at the limit; the active-count assertion drops to 1 and fails if reverted to the blocking drainOne. Matches the sibling pollFreshAccepted. LGTM — thanks!

crypticsaiyan:
Hey @zeshi-du, thanks for the approval. Workflows are still awaiting approval on PRs #6, #7, #9, #22, and #23. Could you approve those runs so CI can finish?

PR #7, opened 2026-06-23
fix(test): use atomic writes for code get --out to prevent file truncation

crypticsaiyan:
## Summary

Previously, running `test code get --out <path>` would open and immediately truncate the destination file **before** issuing the network request. If the request subsequently failed (e.g., network error, 4xx, `NOT_FOUND`) or hit the "no code generated yet" branch (which writes nothing and exits 0), the user's pre-existing file was left completely empty, with no way to recover the original contents.

**Note:** The `bundle.ts` command already correctly mitigates this using an atomic write pattern (writing to a `.tmp` file and renaming it only upon success); `code get --out` had skipped this safety mechanism.

## The Fix

Implemented the atomic write pattern to ensure safe file operations. `openOutputFile` now writes to a sibling temporary file (`.{name}.tmp-{uuid}`) instead of directly targeting the user's requested path.

This ensures `closeOutputFile(sink, commit)` renames the temporary file onto the real `--out` path *only* when content is successfully downloaded and written. On any failure, or if the test has no code yet, the temporary file is safely discarded, and the user's original file is left completely untouched.

## Testing & Validation

**Regression Test:** Added new test cases in `src/commands/test.test.ts` to simulate failure and empty states.  Now, file content is preserved in the following cases (previously emptied):
- **Failing fetch:** Simulated a network error with a pre-existing `--out` file.
- **Empty content:** Simulated a "No code generated yet" response with a pre-existing `--out` file.

- `npm run lint` & `npm run format:check`
- `npm run typecheck`
- `npm test` (1449 tests passing)
- `npm run test:coverage` (≥80% threshold met)

**Files Changed:**
- `src/commands/test.ts`
- `src/commands/test.test.ts`

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -1,4 +1,11 @@
-import { existsSync, mkdirSync, mkdtempSync, readFileSync, writeFileSync } from 'node:fs';
+import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readdirSync,
+  readFileSync,
+  writeFileSync,
+} from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Verified locally — atomic temp-write with rename-on-commit; the original file is left untouched on failure or empty response, matching the existing bundle.ts pattern. One non-blocking nit: the inline-code text path does not set wroteContent=true, so please sanity-check that the empty-file-on-inline case behaves as you intend. LGTM — thanks!

PR #8, opened 2026-06-23
fix(config): treat empty env vars as unset in loadConfig

SahilRakhaiya05:
## Summary

When `TESTSPRITE_API_URL` or `TESTSPRITE_API_KEY` is set but empty/whitespace (e.g. `export TESTSPRITE_API_URL=` in a shell profile), runtime commands no longer resolve to a broken empty endpoint or ignore stored credentials. Matches the normalization already used in `auth configure` / `init`.

**Split from #8** per review - bundle re-commit work moved to a separate PR.

## Changes

- `normalizeEnvVar` exported from `loadConfig` and used for `TESTSPRITE_API_URL` / `TESTSPRITE_API_KEY`
- `auth configure` / `runWhoami` dry-run use the shared helper
- `init` endpoint reporting uses the shared helper

## Testing

- `npx vitest run src/lib/config.test.ts` - 11 passed
- `npx vitest run src/commands/auth.test.ts -t "whitespace|empty / whitespace"` - 4 passed

## Summary by CodeRabbit

- **Bug Fixes**
  - Empty or whitespace-only environment values are now treated as unset, so API URL/API key fallback chains no longer stop at blank strings.
  - Dry-run auth/init flows now resolve to the correct default endpoint when the env URL is blank.
  - HTTP client behavior is corrected so blank API keys trigger the expected “auth required” outcome.

- **New Features**
  - Centralized environment-variable normalization is now used across configuration and command flows.

- **Tests**
  - Expanded coverage for blank API URL/API key handling in config, auth, and client creation.

### DIFF
--- a/src/commands/auth.test.ts
+++ b/src/commands/auth.test.ts
@@ -850,6 +850,21 @@ describe('runWhoami', () => {
     expect(printed).toEqual(sampleMe);
   });
 
+  it('dry-run: whitespace-only TESTSPRITE_API_URL falls through to prod default endpoint', async () => {
+    const { capture, deps } = makeCapture();
+    await runWhoami(
+      { profile: 'default', output: 'text', debug: false, dryRun: true },
+      {
+        ...deps,
+        env: { TESTSPRITE_API_URL: '   ' },
+        credentialsPath,
+        fetchImpl: makeFetch(meResponse()),
+      },
+    );
+    const out = capture.stdout.join('\n');
+    expect(out).toContain('endpoint: https://api.testsprite.com');
+  });
+
   it('L1788: text output includes the resolved endpoint URL', async () => {
     writeProfile(
       'default',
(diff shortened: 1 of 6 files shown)


zeshi-du:
@coderabbitai review

SahilRakhaiya05:
This PR is solving Issue #97

zeshi-du:
Two updates after a deeper review:

1. **The empty-env-var normalization half is good to go.** Please split it into its own PR and it'll merge fast.

2. **The bundle re-commit half needs a rethink after #162** (merging today): `--out` may point at a directory that also holds the user's unrelated files, and #162 establishes the contract that the commit sweep must never touch entries the bundle format doesn't own (`isBundleOwnedEntry`). A whole-directory swap replaces the entire `--out` dir, so the user's foreign files would be evicted along with the prior bundle — silent data loss.

The atomicity goal itself is right (today a failed re-commit can delete `meta.json` without landing a replacement — worth fixing). A compatible shape would be: stage in `.tmp/` as you do now, then swap **only bundle-owned entries** with per-entry aside + rollback — or stage the whole bundle under one owned subdirectory so a single rename swaps it. Happy to review either design.

zeshi-du (approved):
Reviewed: scoped, tests included, CI green across the matrix.

PR #9, opened 2026-06-23
fix(test): dedupe and serialize chunked batch-rerun dispatch

crypticsaiyan:
## Summary

Previously, running `test rerun --all` would split requests into chunks of 50 tests and dispatch all chunks concurrently using `Promise.all`. However, the backend's producer/teardown deduplication logic operates **within a single request**, not across concurrent requests. If two tests in different chunks shared a project's producer or teardown, the backend would independently decide "not yet triggered" for both, causing the shared dependency to execute multiple times.

**Note:** This concurrent dispatch bug applied to both the initial batch dispatch and the deferred-retry loop.

## The Fix

Implemented sequential chunk dispatch and a client-side deduplication defense layer to prevent double-execution:

* **Sequential Dispatch:** Replaced `Promise.all` with sequential execution (one `await` per chunk). By the time chunk N is sent, chunk N-1 has already landed server-side, allowing the backend to correctly deduplicate shared dependencies against the previous chunk.
* **Client-Side Defense:** Added `dedupeBatchRerunAccepted` to drop duplicate `testId` entries from the aggregated `accepted` list, and `mergeBatchRerunClosureByProject` to safely merge `closure.byProject` entries without double-counting. The CLI now emits a warning to `stderr` if it detects and filters out a duplicate trigger.
* **Fixture Update:** Fixed a pre-existing test fixture (`D3: batch rerun summary...`) that was artificially relying on the old double-counting bug.

## Testing & Validation

**Regression Test:** Added new test cases in `src/commands/test.rerun.spec.ts` to verify serialization and deduplication.
* **Concurrency limit:** Simulated timed mock responses to prove that chunks are strictly dispatched sequentially (concurrency never exceeds 1).
* **Client deduplication:** Simulated a backend incorrectly accepting a shared producer across two chunks to assert that the CLI successfully deduplicates the entry and emits a `stderr` warning.

* `npm run lint` & `npm run format:check`
* `npm run typecheck`
* `npm test` (1449 tests passing)
* `npm run test:coverage` (≥80% threshold met; actual ~87%)

**Files Changed:**
* `src/commands/test.ts`
* `src/commands/test.rerun.spec.ts`

### DIFF
--- a/src/commands/test.rerun.spec.ts
+++ b/src/commands/test.rerun.spec.ts
@@ -2431,6 +2431,159 @@ describe('[fix-D] --all resolves >50 tests: chunked batch requests, aggregated r
     // Exactly 50 → 1 request only
     expect(batchCallCount).toBe(1);
   });
+
+  // Regression: chunked batch-rerun dispatched chunks via Promise.all, so
+  // when --all resolves >50 tests every chunk's request was in flight at
+  // once. BE producer/teardown closure dedup happens per-request, so two
+  // concurrent chunks sharing a project's producer could each independently
+  // trigger it. Chunks must be dispatched strictly one at a time.
+  it('60 tests → 2 chunks are dispatched sequentially, not concurrently', async () => {
+    const creds = makeCreds();
+    const allTests = Array.from({ length: 60 }, (_, i) => ({
+      ...FE_TEST,
+      id: `test_seq_${String(i).padStart(3, '0')}`,
+    }));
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Verified locally — sequential chunk dispatch (both the initial and deferred-retry paths) plus client-side dedupe; the concurrency assertion fails if reverted to Promise.all. Great catch on the shared producer/teardown double-execution. LGTM — thanks!

PR #11, opened 2026-06-23
feat(cli): add runtime Node.js version check with clear error message

lxcario:
When run on Node < 20, the CLI crashes with cryptic ESM/syntax errors. This PR adds a top-of-file version guard in `src/index.ts` that checks `process.versions.node` and exits with a clear, helpful message if the major version is below 20.

## Changes

- Added version guard at the top of `src/index.ts` (before any imports that use modern syntax)
- Exits with a clear error message pointing users to https://nodejs.org
- Added `src/version-guard.test.ts` with 8 tests verifying the guard logic

## Testing

- `npm run typecheck` - clean
- `npm run lint:fix` - clean
- `npx vitest run src/version-guard.test.ts` - 8/8 passing

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an early startup Node.js runtime compatibility check; the CLI now exits immediately with code `1` and a clear upgrade message when Node.js is below 20 (including the detected version and an LTS install link).
  * Avoids blocking startup when the Node.js version string can’t be parsed.
* **Tests**
  * Added automated coverage for version parsing and the supported/unsupported Node.js decision logic, including boundary and unparseable input cases.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,15 +1,5 @@
 #!/usr/bin/env node
 
-// Guard: exit early with a clear message on unsupported Node.js versions.
-// This runs before any ESM/modern-syntax imports that would crash with cryptic errors.
-const major = Number(process.versions.node.split('.')[0]);
-if (major < 20) {
-  process.stderr.write(
-    `Error: testsprite requires Node.js >= 20 (found ${process.versions.node}).\nInstall the latest LTS from https://nodejs.org\n`,
-  );
-  process.exit(1);
-}
-
 import { Command, CommanderError } from 'commander';
 import { createAgentCommand } from './commands/agent.js';
 import { createAuthCommand } from './commands/auth.js';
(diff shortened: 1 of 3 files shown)


zeshi-du:
@coderabbitai review

lxcario:
Addressed the earlier review signals. The test previously re-declared the guard functions inside the spec (hence the trivial-assertion flag) -- it wasn't exercising the real code. I extracted the logic into a documented module, src/version-guard.ts (parseMajorVersion, shouldRejectNodeVersion, MIN_SUPPORTED_NODE_MAJOR, all with JSDoc), wired src/index.ts to use it, and rewrote src/version-guard.test.ts to import the real functions. Coverage now includes the exact-floor boundary (20 supported, 19 rejected) and malformed-input behavior. typecheck, build, lint and prettier are clean; node dist/index.js --version still runs. @coderabbitai review

lxcario:
Rebased onto current main; all 8 checks are green. The guard logic lives in src/version-guard.ts with real exports the tests import, which addresses the earlier trivial-assertion note. On the optional suggestion to run the version check before the commander import: happy to move it if you'd prefer, but I left it as-is since it's non-blocking and commander targets broad Node compatibility. Ready when you are.

zeshi-du (approved):

PR #12, opened 2026-06-23
feat(cli): respect NO_COLOR environment variable per no-color.org

lxcario:
The CLI uses ANSI escape codes for the TTY progress ticker (`\x1b[2K\r`). The [NO_COLOR standard](https://no-color.org/) says tools should suppress color/ANSI output when `NO_COLOR` env is set.

## Changes

- Added `isNoColor()` utility in `src/lib/ticker.ts` that checks for `NO_COLOR` in the environment
- When `NO_COLOR` is set and the terminal is a TTY, the ticker emits plain-text lines via `stderrWrite` instead of using ANSI escape sequences via `rawWrite`
- Added a 4th optional `noColor` parameter to `createTicker()` (defaults to env detection)
- Added `NO_COLOR` to the environment variables table in DOCUMENTATION.md
- Added 5 new tests (3 for NO_COLOR ticker behavior + 2 for `isNoColor` utility)

## Testing

- `npm run typecheck` - clean
- `npm run lint:fix` - clean
- `npx vitest run src/lib/ticker.spec.ts` - 19/19 passing (all existing + 5 new)

## Summary by CodeRabbit

* **New Features**
  * Added `NO_COLOR` support to disable ANSI escape sequences for the TTY progress ticker.
  * When color is disabled, ticker updates/finalization emit plain, timestamped text (no in-place ANSI overwrite behavior).

* **Documentation**
  * Updated the environment variables reference to include `NO_COLOR` and improved table formatting.

* **Tests**
  * Expanded ticker test coverage to validate `NO_COLOR` parsing and verify plain-text output behavior in TTY mode.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -426,12 +426,13 @@ These apply to every command:
 
 ### Environment variables
 
-| Variable                        | Purpose                                                                           |
-| ------------------------------- | --------------------------------------------------------------------------------- |
-| `TESTSPRITE_API_KEY`            | API key — overrides the credentials file                                          |
-| `TESTSPRITE_API_URL`            | API endpoint — overrides the credentials file                                     |
-| `TESTSPRITE_PROFILE`            | Active profile (below `--profile`, above `default`)                               |
-| `TESTSPRITE_REQUEST_TIMEOUT_MS` | Per-request timeout in **milliseconds** (default `120000`, range `1000`–`600000`) |
+| Variable                        | Purpose                                                                                 |
+| ------------------------------- | --------------------------------------------------------------------------------------- |
+| `TESTSPRITE_API_KEY`            | API key — overrides the credentials file                                                |
+| `TESTSPRITE_API_URL`            | API endpoint — overrides the credentials file                                           |
+| `TESTSPRITE_PROFILE`            | Active profile (below `--profile`, above `default`)                                     |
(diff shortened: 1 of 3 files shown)


zeshi-du:
@coderabbitai review

zeshi-du:
Thanks @lxcario — this is really close and on-pattern! 🙌

One small thing before we merge: `isNoColor()` uses `'NO_COLOR' in env`, which treats `NO_COLOR=''` (empty) as *disabling* color — but the [no-color.org](https://no-color.org) spec (and your own doc comment) require the variable to be **non-empty** to take effect. Could you switch it to check for a non-empty value, and flip the matching test so `isNoColor({ NO_COLOR: '' })` expects `false`? CodeRabbit flagged the same point.

Once that's in, we'll take it. Thanks for the contribution! 🚀

lxcario:
Thanks @zeshi-du! Addressed the feedback: isNoColor now requires a non-empty NO_COLOR value (per no-color.org), and I flipped the test so isNoColor({ NO_COLOR: '' }) expects false. ticker.spec.ts is green (19/19), typecheck and prettier clean. Pushed -- ready for another look.

lxcario:
Rebased onto current main; all 8 checks are green. isNoColor() now requires a non-empty NO_COLOR value per no-color.org, and the test expects isNoColor({ NO_COLOR: '' }) to be false, as requested. Ready for merge.

zeshi-du (approved):

PR #17, opened 2026-06-24
fix(cli): validate --request-timeout flag and de-duplicate its parser

### LINKED ISSUE #16: Invalid `--request-timeout` value is silently ignored (falls back to 120s); parser duplicated 5×
## Summary

Two related problems with the global `--request-timeout <seconds>` flag:

### 1. Invalid values are silently dropped

`parseRequestTimeoutFlag` returns `undefined` for any non-numeric / zero / negative input. `undefined` means "flag not supplied", so the factory falls back to `TESTSPRITE_REQUEST_TIMEOUT_MS` or the 120s default. A natural typo like `--request-timeout 30s` (meaning "30 seconds") therefore runs the command with the **default 120s deadline** while the operator believes they set 30s — no error, no signal.

```console
$ testsprite --request-timeout 30s project list   # runs with 120s, exits normally
```

Every other validated flag in the CLI fails loudly on bad input — `--page-size 0`, `--page-size 101`, `--output yaml`, `--type junk` all exit 5 with a `VALIDATION_ERROR`. `--request-timeout` is the odd one out.

### 2. The parser is duplicated five times

The exact same `parseRequestTimeoutFlag` body is copy-pasted into `auth.ts`, `project.ts`, `usage.ts`, `init.ts`, and `test.ts`. Drift between copies would silently change timeout behaviour depending on which command you ran.

## Suggested fix

Hoist a single `parseRequestTimeoutFlag` into `src/lib/client-factory.ts` (next to `resolveRequestTimeoutMs` and the `REQUEST_TIMEOUT_*` constants) and make the **flag** strict: a non-numeric / zero / negative value throws a typed `VALIDATION_ERROR` (exit 5). Positive out-of-range values keep flowing through to `resolveRequestTimeoutMs`, which already clamps to `[1s, 600s]`. The `TESTSPRITE_REQUEST_TIMEOUT_MS` **env-var** path stays lenient (a stray global env var should not hard-fail every command) — only the explicit per-invocation flag becomes strict.

I have a PR ready for this.

Davidson3556:
Fixes #16

Two related problems with the global `--request-timeout <seconds>` flag:

1. **Invalid values were silently dropped.** `parseRequestTimeoutFlag` returned `undefined` for any non-numeric / zero / negative input, and `undefined` is indistinguishable from "flag omitted" — so the command fell back to `TESTSPRITE_REQUEST_TIMEOUT_MS` or the 120s default. A natural typo like `--request-timeout 30s` (meaning "30 seconds") ran with the **default 120s deadline** while the operator believed they had set 30s, with no error. Every other validated flag (`--page-size`, `--output`, `--type`) exits 5 on bad input; this one was the exception.

2. **The parser was duplicated five times.** The identical `parseRequestTimeoutFlag` body lived in `auth.ts`, `project.ts`, `usage.ts`, `init.ts`, and `test.ts`. Drift between copies would silently change timeout behaviour depending on the command.

Changes:
- Hoist one `parseRequestTimeoutFlag` into `src/lib/client-factory.ts`, co-located with `resolveRequestTimeoutMs` and the `REQUEST_TIMEOUT_*` constants; the five command files import it.
- Make the **flag** strict: a non-numeric / zero / negative value throws a typed `VALIDATION_ERROR` (exit 5).
- Preserve intentional leniency where it exists: positive out-of-range values still flow through to `resolveRequestTimeoutMs`, which already clamps to `[1s, 600s]`; the `TESTSPRITE_REQUEST_TIMEOUT_MS` **env-var** path stays lenient (its existing tests assert non-numeric/zero/negative env vars fall back to the default) — a stray global env var should not hard-fail every command.
- Add unit coverage for `parseRequestTimeoutFlag` and a subprocess regression that `--request-timeout 30s` exits 5 instead of silently using 120s.

Before (main) — the invalid flag is ignored and execution proceeds past validation (here reaching the auth step), so it would have silently run with the default 120s:

image

After (this branch) — the flag is validated up front; valid values still proceed:

  image

Tests:

image

---

Problem: an explicit `--request-timeout` value that didn't parse was silently discarded, so the operator's chosen deadline was ignored without any signal — and the parser was copy-pasted into five files.

Alternatives considered:
1. Keep the lenient behaviour and just de-duplicate. Rejected: that leaves the real footgun (a typo silently runs at 120s) in place.
2. Make both the flag and the env var strict. Rejected: the env-var leniency is intentional and pinned by `resolveRequestTimeoutMs` tests — a globally-set `TESTSPRITE_REQUEST_TIMEOUT_MS` should not hard-fail every single command. Only the explicit per-invocation flag should fail loudly.
3. (chosen) One shared parser, strict on the flag, lenient on the env var, with out-of-range positives still clamped downstream.

Key function: `parseRequestTimeoutFlag(raw)` — `undefined` → `undefined` (flag omitted, factory falls back to env/default); a positive number → seconds×1000 ms (including out-of-range values, which `resolveRequestTimeoutMs` clamps to `[1s, 600s]`); non-numeric / zero / negative → throws the typed `VALIDATION_ERROR` (exit 5). Placed in `client-factory.ts` next to the related timeout logic so the flag, env, and default resolution all live together.

Edge cases handled/tested: omitted flag; positive seconds (incl. fractional like `2.5`); positive out-of-range (`700` parses to 700000, clamped later, not rejected); `abc`, `30s`, `0`, `-5`, `NaN`, `Infinity`, and empty string each throw exit 5; subprocess test confirms the CLI exit code is 5, not a silent fall-back.

---

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `--request-timeout` handling: invalid or non-positive values now fail immediately with a `VALIDATION_ERROR` (exit code 5) instead of silently falling back to defaults.
  * Standardized request-timeout parsing across multiple CLI commands, including `--wait` behavior.

* **Tests**
  * Added unit tests for request-timeout parsing (valid conversions, omitted flag, and invalid inputs with expected error details).
  * Added a subprocess end-to-end test verifying the CLI rejects `--request-timeout 30s` in JSON mode with the correct error envelope and next action.

### DIFF
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -2,6 +2,7 @@ import { Command } from 'commander';
 import {
   emitDryRunBanner,
   makeHttpClient,
+  parseRequestTimeoutFlag,
   type CommonOptions as FactoryCommonOptions,
 } from '../lib/client-factory.js';
 import type { ErrorCode } from '../lib/errors.js';
@@ -325,19 +326,6 @@ function resolveCommonOptions(command: Command): CommonOptions {
     dryRun: globals.dryRun ?? false,
     requestTimeoutMs: parseRequestTimeoutFlag(globals.requestTimeout),
   };
-}
-
-/**
- * Parse the `--request-timeout <seconds>` flag value into milliseconds.
- * Returns `undefined` when the flag was not supplied (factory falls back to
- * the env var / default). Silently clamps out-of-range values — the
- * factory applies the same clamp so there is no double-clamp risk.
- */
-function parseRequestTimeoutFlag(raw: string | undefined): number | undefined {
-  if (raw === undefined) return undefined;
-  const n = Number(raw);
-  if (!Number.isFinite(n) || n <= 0) return undefined;
-  return Math.round(n * 1000); // seconds → milliseconds
 }
 
 function makeOutput(mode: OutputMode, deps: AuthDeps): Output {
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -11,7 +11,10 @@
  */
 
 import { Command } from 'commander';
-import type { CommonOptions as FactoryCommonOptions } from '../lib/client-factory.js';
+import {
+  parseRequestTimeoutFlag,
+  type CommonOptions as FactoryCommonOptions,
+} from '../lib/client-factory.js';
 import { emitDeprecationNotice } from '../lib/deprecate.js';
 import { CLIError } from '../lib/errors.js';
 import { GLOBAL_OPTS_HINT, Output } from '../lib/output.js';
@@ -431,13 +434,6 @@ function resolveCommonOptions(command: Command): CommonOptions {
     dryRun: globals.dryRun ?? false,
     requestTimeoutMs: parseRequestTimeoutFlag(globals.requestTimeout),
   };
-}
-
-function parseRequestTimeoutFlag(raw: string | undefined): number | undefined {
-  if (raw === undefined) return undefined;
-  const n = Number(raw);
-  if (!Number.isFinite(n) || n <= 0) return undefined;
-  return Math.round(n * 1000);
 }
 
 const SETUP_DESCRIPTION =
--- a/src/commands/project.ts
+++ b/src/commands/project.ts
@@ -4,6 +4,7 @@ import { Command } from 'commander';
 import {
   emitDryRunBanner,
   makeHttpClient,
+  parseRequestTimeoutFlag,
   type CommonOptions as FactoryCommonOptions,
 } from '../lib/client-factory.js';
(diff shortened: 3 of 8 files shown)


zeshi-du:
@coderabbitai review

Davidson3556:
Thanks! Addressed the CodeRabbit nitpick — `parseRequestTimeoutFlag` now surfaces the offending raw value in the error message (e.g. `Flag \`--request-timeout\` is invalid: "30s" is not valid — must be a positive number of seconds.`), matching the pattern `assertValidEndpointUrl` already uses in this file. Also rebased onto the latest `main` (v0.2.0). All gates green locally (typecheck, lint, format, unit 1579, e2e).

zeshi-du (approved):

PR #21, opened 2026-06-24
fix(cli): validate profile names to prevent credentials-file corruption

### LINKED ISSUE #20: A malformed `--profile` name silently corrupts ~/.testsprite/credentials (key not persisted)
## Summary

The profile name (`--profile` / `TESTSPRITE_PROFILE`) is written verbatim as an INI section header (`[name]`) in `~/.testsprite/credentials`, but is never validated. A name containing characters that break the INI grammar corrupts the file or silently fails to persist the credential.

## Reproduction (round-trip of the pure serialize/parse helpers)

```js
import { serializeCredentials, parseCredentials } from '@testsprite/testsprite-cli/dist/lib/credentials.js';

const file = { 'prod]': { apiKey: 'sk-123', apiUrl: 'https://api.testsprite.com' } };
console.log(serializeCredentials(file));
// [prod]]
// api_key = sk-123
// api_url = https://api.testsprite.com
console.log(Object.keys(parseCredentials(serializeCredentials(file))));
// []   <-- the section AND its api_key/api_url lines are silently dropped
```

So `testsprite setup --profile "prod]"` would report success while the key never persists (the next command can't read it back). Other corrupting inputs:

- a newline in the name splits the header across two lines;
- leading/trailing whitespace doesn't round-trip (the parser trims section names, so `[ prod ]` reads back as `prod`).

## Suggested fix

Validate the profile name against a conservative allowlist (letters, digits, dot, underscore, hyphen — covering `default`, `prod`, `ci-staging`, `team.qa`) at the profile-keyed credential entry points (`readProfile` / `writeProfile` / `deleteProfile`), throwing a typed `VALIDATION_ERROR` (exit 5) before any file write.

I have a PR ready for this.

Davidson3556:
Fixes #20

A profile name (`--profile` / `TESTSPRITE_PROFILE`) is written verbatim as an INI section header (`[name]`) in `~/.testsprite/credentials`, but was never validated. A name containing the characters that break that grammar silently corrupted the file:

- `--profile "prod]"` → serialises to `[prod]]`, which the section regex cannot match, so the `api_key`/`api_url` lines that follow are **dropped on read**. `setup` reports success while the credential never persists.
- a newline in the name splits the header across two lines.
- leading/trailing whitespace doesn't round-trip (the parser trims section names, so `[ prod ]` reads back as `prod`).

Changes:
- Add `assertValidProfileName` in `credentials.ts` — a conservative allowlist (letters, digits, dot, underscore, hyphen) that covers conventional names (`default`, `prod`, `ci-staging`, `team.qa`) and cannot corrupt the INI file.
- Call it from every profile-keyed entry point: `readProfile`, `writeProfile`, `deleteProfile`. A malformed name now throws a typed `VALIDATION_ERROR` (exit 5) **before any file write**.
- Add unit coverage for the guard and the three entry points, plus a subprocess regression.

Before (main) — the round-trip silently drops the credential (`prod]` → no readable profile):

    image

After (this branch) — rejected up front; valid names unaffected:

image
---

Problem: the profile name is interpolated straight into an INI section header during serialization (`[${section}]`), but the parser only accepts `^\[([^\]]+)\]$` and trims the captured name. So any name containing `]`, a newline, or surrounding whitespace either fails to match (dropping the section and its key/url lines) or doesn't round-trip — and `setup` reports success regardless, leaving the user with credentials that can't be read back.

Alternatives considered:
1. Escape/quote profile names in the INI writer and teach the parser to unescape. Rejected: it complicates the file format, and there's no value in supporting exotic profile names — they're short identifiers.
2. Validate only at write time (`writeProfile`). Rejected: reading with a malformed name should also fail clearly rather than silently returning "no profile", so the guard belongs on the read path too.
3. (chosen) A small allowlist validator called from all three profile-keyed entry points (`readProfile` / `writeProfile` / `deleteProfile`), throwing the same typed `VALIDATION_ERROR` (exit 5) used elsewhere. This catches the bad name on every credential flow (setup, status, logout, and any command that resolves config) before a write can corrupt the file.

Key function: `assertValidProfileName(profile)` tests against `^[A-Za-z0-9._-]+$` and throws `localValidationError('profile', …, 'flag')` otherwise. Placed in `credentials.ts` next to the parse/serialize logic that defines the constraint (no import cycle — `errors.ts` does not import `credentials.ts`).

Edge cases handled/tested: accepts `default`, `prod`, `ci-staging`, `team.qa`, `a_b`, `P1`; rejects `prod]`, `[weird`, embedded newline, whitespace-padded, names with spaces or `/`, and the empty string; `writeProfile` rejects before creating the file (verified the file is not created).

---

---

### DIFF
--- a/src/lib/credentials.test.ts
+++ b/src/lib/credentials.test.ts
@@ -4,6 +4,7 @@ import { join } from 'node:path';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import {
   DEFAULT_PROFILE,
+  assertValidProfileName,
   defaultCredentialsPath,
   deleteProfile,
   ensureRestrictiveMode,
@@ -13,6 +14,7 @@ import {
   serializeCredentials,
   writeProfile,
 } from './credentials.js';
+import { ApiError } from './errors.js';
 
 let tmpRoot: string;
 let credentialsPath: string;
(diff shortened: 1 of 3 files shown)


zeshi-du (dismissed):
Verified locally — the allowlist check is wired into all three profile entry points before any write, so a malformed name can no longer corrupt the credentials file. The new tests fail if the guard is disabled. LGTM — thanks!

PR #22, opened 2026-06-24
fix(cli): emit JSON envelope for Commander parse errors under --output json

crypticsaiyan:
## Summary

When `--output json` is set, Commander-level parse errors (e.g., unknown command, missing arg) previously exited with code `5` but wrote plain text to `stderr`. This broke machine consumers expecting structured JSON data. All other error paths already correctly emit JSON; Commander was the sole exception.

## The Fix

**Root Cause:** The `configureOutput` setting was only applied to the root program. Subcommands kept Commander's default behavior, which writes plain text immediately before the `CommanderError` is thrown and the output mode is checked.

**Changes:**
* **Propagation & Buffering:** `applyExitOverrideDeep` now passes `configureOutput` to all subcommands. Errors are buffered instead of written immediately.
* **Smart Envelope Generation:** The `CommanderError` catch block now evaluates the output mode to write either a `VALIDATION_ERROR` JSON envelope or standard plain text.
* **Argv Scan Fallback:** Added a fallback to detect `--output json` if it appears in the command line after the invalid argument.
* **Testability:** Extracted `renderCommanderError` to `src/lib/render-error.ts` for direct unit testing without needing a subprocess.

## Testing & Validation

**Regression Tests:** 
*  **Unit Tests:** 8 tests added in `src/lib/render-error.test.ts` covering JSON/text modes, null fallbacks, and envelope shapes.
* **Subprocess Tests:** 4 tests added in `test/cli.subprocess.test.ts` covering missing args, unknown commands, `argv` fallbacks, and text mode baselines.

* Lint & format check
* TypeScript compiler (`typecheck`)
* Unit tests (1459 passing)
* Coverage (≥80% threshold met)

**Files Changed:**
* `src/lib/render-error.ts`
* `src/index.ts`
* `src/lib/render-error.test.ts`
* `test/cli.subprocess.test.ts`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,7 +12,7 @@ import { createTestCommand } from './commands/test.js';
 import { createUsageCommand } from './commands/usage.js';
 import { ApiError, CLIError, RequestTimeoutError } from './lib/errors.js';
 import { Output, isOutputMode } from './lib/output.js';
-import { rephraseUnknownOption } from './lib/render-error.js';
+import { renderCommanderError, rephraseUnknownOption } from './lib/render-error.js';
 import { maybeEmitSkillNudge } from './lib/skill-nudge.js';
 import { VERSION } from './version.js';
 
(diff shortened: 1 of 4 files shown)


zeshi-du (approved):
Verified locally — configureOutput now propagates to subcommands and the Commander error path emits a proper JSON envelope (confirmed an unknown command with --output json returns valid JSON and exit 5). LGTM — thanks!

PR #23, opened 2026-06-24
fix(bundle): add retry loop to streamUrlToFile for transport failures

crypticsaiyan:
## Summary
`streamUrlToFile` promised a "bounded retry on transport failures" in its JSDoc but made only one attempt. Transient network blips during downloads caused hard failures, forcing manual re-runs.

## The Fix
* **Bounded Retry Loop:** Added a 3-attempt loop (`STREAM_URL_MAX_RETRIES`) with a 1s delay for transport or mid-stream pipeline errors.
* **Smart Fail-Fast:** HTTP 4xx/5xx errors (`ApiError`) skip retries and fail immediately (e.g., expired presigned URLs will not recover).
* **Testability:** Added an optional `deps.sleep` injection for deterministic testing without real-world delays.

## Testing
Added 5 tests to `src/lib/bundle.test.ts` verifying:
* First-attempt success
* Transient recovery (retry-then-succeed)
* Retry exhaustion (`TransportError`)
* Fast failure on HTTP errors (`ApiError`)
* Correct injected `sleep` calls between attempts

**Files Changed:**
* `src/lib/bundle.ts`
* `src/lib/bundle.test.ts`

### DIFF
--- a/src/lib/bundle.test.ts
+++ b/src/lib/bundle.test.ts
@@ -8,6 +8,9 @@
  * the full http+fetch path is wired against MSW).
  */
 
+import { mkdtempSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
 import { describe, expect, it } from 'vitest';
 import {
   applyFailedOnly,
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Verified locally — bounded retry on transport errors, fast-fail on 4xx/5xx, and the injected sleep keeps tests deterministic. The retry-then-succeed test fails if the loop is reverted to a single attempt. LGTM — thanks!

PR #26, opened 2026-06-24
fix(test): validate failure get --out before API fetch

SahilRakhaiya05:
## Summary

`runFailureGet()` had a comment claiming `--out` is validated before the network call, but it only assigned `opts.out` and immediately fetched `GET /tests/{id}/failure`. Invalid paths (empty `--out`, missing parent directory, etc.) wasted an API call before failing with exit 5.

`runArtifactGet()` and `runCodeGet()` already fast-fail on bad `--out` paths first; this aligns `failure get`.

## The Fix

When `--out` is provided:
1. `resolveBundleDir(opts.out)` - rejects empty paths
2. `assertOutDirParentExists(resolvedDir)` - rejects missing parents / file targets
3. Only then issue the API fetch and `writeBundle`

## Testing

- Added 2 regression tests in `src/commands/test.test.ts` (zero fetch calls on invalid `--out`)
- `npx vitest run src/commands/test.test.ts -t runFailureGet` - 15 passed

**Files changed:**
- `src/commands/test.ts`
- `src/commands/test.test.ts`

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -3220,6 +3220,52 @@ function makeFailureContext(overrides: Partial<CliFailureContext> = {}): CliFail
 }
 
 describe('runFailureGet', () => {
+  it('--out rejects an empty path with VALIDATION_ERROR (exit 5) before any network I/O', async () => {
+    const { credentialsPath } = makeCreds();
+    let fetchCalls = 0;
+    const fetchImpl = makeFetch(() => {
+      fetchCalls += 1;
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Verified locally — the pre-fetch `resolveBundleDir` + `assertOutDirParentExists` now mirror `runArtifactGet`/`runCodeGet`, and the two new tests genuinely fail when the guard is removed. LGTM — thanks!

PR #31, opened 2026-06-25
fix(cli): route interactive prompts and prelude to stderr (keep stdout pure)

### LINKED ISSUE #30: Interactive prompts and the configure prelude write to stdout, breaking --output json purity
## Summary

The interactive prompts in `prompt.ts` (the API-key prompt in `setup`/`auth configure`, the target prompt in `agent install`) write the question and masking characters to **stdout**, and the `Configuring profile …` prelude in `auth.ts` defaults to stdout as well.

On any interactive path this mixes UI text into stdout. Under `--output json` it breaks the contract that stdout is a single JSON document — a machine consumer doing `JSON.parse(stdout)` fails on the prompt text. The repo already enforces "stdout is a pure result stream" (the `stdoutPurity` helper, §8.1), but the interactive prompt path was missed.

## Suggested fix

Default prompts and the informational prelude to **stderr** — they're interactive UI, not result data. stderr is still the user's TTY, so prompts stay visible; stdout carries only the result. Callers that inject explicit streams are unaffected.

I have a PR ready for this.

Davidson3556:
Fixes #30

Interactive prompts wrote to **stdout**, polluting the result stream:

- `prompt.ts` (`promptText` / `promptSecret`) defaulted the question + masking to stdout — used by the API-key prompt in `setup`/`auth configure` and the target prompt in `agent install`.
- the `Configuring profile …` prelude in `auth.ts` defaulted to stdout.

On any interactive path this mixes UI text into stdout, and under `--output json` it breaks the contract that stdout is a single JSON document (a consumer doing `JSON.parse(stdout)` fails on the prompt text). The repo already enforces stdout purity elsewhere (the `stdoutPurity` helper, §8.1) — the interactive prompt path was the gap.

Change: default both to **stderr**. Prompts and informational preludes are interactive UI, not result data; stderr is still the user's TTY so they stay visible, and the secret is still never echoed. Callers that inject explicit streams are unaffected.

The cleanest reproduction uses `agent install` with no `--target` (it prompts, and is fully local — no key/network). Redirect **stdout** to a file; the prompt should appear on your terminal (stderr) and the file should hold only the result.

image

---

Problem: the repo treats stdout as a pure result stream (so `--output json` yields one parseable document and text-mode stdout is script-clean), but interactive prompts and the configure prelude wrote to stdout, violating that on the interactive path.

Approach: prompts are UI, not data, so the fix is to default their output stream to stderr. `prompt.ts` already accepts an injectable `output` stream; I only changed the default (`process.stdout` → `process.stderr`) in `promptText`/`promptSecret`, and changed the `auth.ts` prelude's default writer the same way. stderr is the controlling terminal in an interactive session, so the prompt stays visible; nothing about masking or the "never echo the secret" behavior changes.

Alternatives considered: (1) gate the stream on `--output json` only — rejected, because text-mode stdout should be clean too (e.g. `KEY=$(testsprite … )`), and a mode-dependent stream is more surprising than "prompts always go to stderr," which is the conventional behavior for CLIs. (2) Strip the prompt from stdout after the fact — not possible/clean.

Edge cases handled/tested: a `promptText` regression test confirms the question goes to stderr (not stdout) when no stream is injected; an `auth` regression test confirms the `Configuring profile …` prelude lands on stderr while the result stays on stdout. Existing prompt tests (which inject an explicit `output` stream) are unaffected, and the secret-not-echoed test still passes.

Note on the tests: I swap the global `process.stderr`/`process.stdout` writers manually (save/restore) rather than `vi.spyOn`, because `vi.spyOn(process.stderr, 'write')` did not intercept the writes in this setup — the manual override is reliable since the code reads `process.stderr` at call time.

---

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Interactive prompts and “Configuring profile …” prelude messages now display on `stderr` by default instead of `stdout`.
  * Keeps `stdout` clean for structured output, including JSON mode.
  * Added regression tests covering prompt text, secret input behavior, and configure prelude routing.

### DIFF
--- a/src/commands/auth.test.ts
+++ b/src/commands/auth.test.ts
@@ -140,6 +140,35 @@ describe('runConfigure', () => {
     expect(capture.prelude.join('')).toContain('Configuring profile "default"');
   });
 
+  it('routes the interactive prelude to stderr by default, keeping stdout for the result', async () => {
+    // Regression: the prelude used to default to process.stdout, polluting the
+    // result stream (and the JSON document under --output json). With no
+    // injected preludeWrite/stderr, the default must land on stderr, not stdout.
+    const stdout: string[] = [];
+    const errChunks: string[] = [];
+    const origErr = process.stderr.write.bind(process.stderr);
+    (process.stderr as unknown as { write: (c: string) => boolean }).write = c => {
+      errChunks.push(String(c));
+      return true;
+    };
+    try {
+      await runConfigure(
+        { profile: 'default', output: 'text', debug: false, fromEnv: false },
+        {
+          stdout: line => stdout.push(line),
+          prompt: { secret: vi.fn(async () => 'sk-typed') },
+          fetchImpl: meOkFetch,
+          credentialsPath,
+          env: {},
+        },
+      );
+    } finally {
+      (process.stderr as unknown as { write: typeof origErr }).write = origErr;
+    }
+    expect(errChunks.join('')).toContain('Configuring profile "default"');
+    expect(stdout.join('\n')).not.toContain('Configuring profile');
+  });
+
   it('interactive path resolves the endpoint from TESTSPRITE_API_URL without prompting', async () => {
     const { capture, deps } = makeCapture();
     const prompt = { secret: vi.fn(async () => 'sk-typed') };
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -73,7 +73,10 @@ export async function runConfigure(opts: ConfigureOptions, deps: AuthDeps = {}):
   const env = deps.env ?? process.env;
   const credentialsPath = deps.credentialsPath ?? defaultCredentialsPath();
   const out = makeOutput(opts.output, deps);
-  const prelude = deps.preludeWrite ?? ((chunk: string) => process.stdout.write(chunk));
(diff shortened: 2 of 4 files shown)


zeshi-du:
@coderabbitai review

Davidson3556:
Thanks for the review! Addressed the CodeRabbit feedback and the failing Lint check:

1. **Lint fix (the ❌ check):** removed the now-unused `vi` import from `prompt.test.ts` — that was the ESLint `no-unused-vars` failure.
2. **Added the missing `promptSecret` regression test** for its default-to-stderr routing (previously only `promptText` was covered).

I intentionally did **not** switch to `vi.spyOn(process.stderr, 'write')` as suggested: in this Vitest setup the spy does not intercept `process.stderr.write` (the call count comes back 0), so the test would silently pass without actually asserting anything. The manual save/restore of the stream writers is reliable because `prompt.ts` reads `process.stderr`/`process.stdout` at call time. Removing the unused `vi` import is the correct resolution here.

Also rebased onto the latest `main` (v0.2.0). All gates green locally (typecheck, lint, format, unit 1571, e2e).

zeshi-du (approved):

PR #4, opened 2026-06-17
fix(test): restore full test suite on Windows

SahilRakhaiya05:
Fixes 10 Windows-only test failures so the Vitest suite runs cleanly on Windows dev machines. Linux/macOS CI behavior is unchanged.

Windows result:

* 1416 passing
* 3 skipped intentionally on Windows only

## What changed

* Fixed `resolveBundleDir` to strip both `/` and `\` trailing separators while preserving roots like `C:\` and `/`.
* Updated subprocess tests to set both `HOME` and `USERPROFILE` so credentials stay inside the test temp directory on Windows.
* Replaced Unix-only `rm` usage with `unlinkSync()`.
* Added `test/helpers/execNpm.ts` so `npm` subprocess calls work on Windows `.cmd` shims.
* Fixed CRLF handling in frontmatter parsing by stripping trailing `\r`.
* Added `.gitattributes` to normalize line endings to LF.
* Skipped symlink tests only when the OS cannot create symlinks.
* Skipped Unix `0600` chmod assertions on Windows because NTFS does not enforce POSIX file modes.
* Replaced Unix-only path assertions with `path.resolve()`, `path.join()`, and `homedir()` based checks.

## Test Plan

* [x] `npm test` — 1416 passed, 3 skipped on Windows only
* [x] `npm run typecheck`
* [x] `npm run lint`
* [x] Subprocess suite — 49/49 passed
* [x] Help snapshot suite — passed

Skipped on Windows by design:

* 2 symlink tests requiring Developer Mode/elevation
* 1 chmod `0600` test due to NTFS behavior

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform behavior in CLI-related tests with a new npm execution helper and isolated home environment settings.
  * Fixed path normalization for bundled directories (trailing separators) and strengthened path/cwd assertions in related tests.
  * Made filesystem-related tests more robust by conditionally skipping symlink safety checks and relaxing Windows-specific permission assertions.
  * Reduced text discrepancies by enforcing LF line endings and stripping trailing CR from parsed frontmatter descriptions.

Solving Issue #108

PR #32, opened 2026-06-25
fix(credentials): serialize profile writes with cross-process file locking

SahilRakhaiya05:
## Summary

writeProfile and deleteProfile perform read-modify-write on `~/.testsprite/credentials`. The final `rename` is atomic, but the read and write are not one operation. Two concurrent CLI processes (e.g. two terminals running `testsprite setup` for different profiles, or `setup` + `auth remove`) can each read the same snapshot; the last rename wins and **silently drops** the other profile update - real credentials data loss.

## The fix

- Wrap `writeProfile` / `deleteProfile` in `withCredentialsLock()`
- Exclusive lock file at `{credentialsPath}.lock` via `openSync(..., 'wx')`
- Stale-lock reclamation when holder pid is dead or lock age > 30s
- Bounded retry (25ms back-off, 10s max wait)

## Tests

- Subprocess regression: parallel writers for `dev` + `staging` profiles - both survive
- Lock file is removed after `writeProfile` / `deleteProfile` complete

## Notes

- Does not overlap with open PRs (#4 Windows tests, #8 config/bundle, #26 failure-get preflight)
- Full `npm test` on Windows still has ~15 pre-existing failures covered by PR #4; new credentials tests pass locally

## Summary

* **Bug Fixes**
  * Prevented concurrent credential updates from clobbering each other by serializing profile writes/deletes with a cross-process lock.
  * Improved reliability of `configure` and `logout` flows to ensure credential persistence/removal completes before reporting results.

* **Tests**
  * Added a concurrency test that performs simultaneous profile writes and verifies both updates persist.
  * Added assertions that lock artifacts are cleaned up after write/delete operations.
  * Updated existing tests to properly await credential operations.

* **Chores**
  * CI now builds before tests, and the test runner avoids parallel workers to reduce flakiness.

Solving Issue #109


zeshi-du:
@coderabbitai review

zeshi-du:
Thanks @SahilRakhaiya05 — strong work here. The cross-process test that spawns real child processes is exactly the rigor we want to see. 🙌

One thing to close before merge: the lock has a create-then-write TOCTOU. `openSync(lockPath, 'wx')` followed by a separate `writeFileSync` of the PID means a contender that hits `EEXIST` in that window can read an empty/partial lock, judge it stale, and barge into the critical section — reintroducing the exact race this fixes. Could you make lock creation atomic — write PID+timestamp to a temp file and `renameSync` it into place (the same pattern `writeCredentialsAtomic` already uses in this file)? CodeRabbit flagged this as Critical as well.

Once that's in, happy to merge. Thanks! 🚀

SahilRakhaiya05:
@zeshi-du  Thanks for the detailed review  I have updated all the requested changes.

zeshi-du:
Substance looks right — the `wx`-flag atomic create + pid/timestamp stale-lock detection is the correct core. One change needed before merge: the busy-wait `syncSleep` (25 ms spin) burns a full CPU core for up to the whole lock-wait window. The `withCredentialsLock` callers are already async, so make the wait a real timer:

```ts
await new Promise(resolve => setTimeout(resolve, LOCK_RETRY_DELAY_MS));
```

If there's a sync-context constraint I'm missing, `Atomics.wait` on a `SharedArrayBuffer` is the non-spinning sync alternative. With the spin removed, this merges.

zeshi-du (changes requested):
The locking design itself looks right now — `wx` lock file with pid+timestamp, stale reclaim via pid-liveness + 30s age, bounded wait, release in `finally`, and the async ripple through `writeProfile`/`deleteProfile` is threaded through every caller. The cross-process child test is a genuinely good test. Two infra-level requests before merge:

1. **Don't serialize the whole unit suite.** `fileParallelism: false` in `vitest.config.ts` applies to all ~1.8k unit tests to fix a race between the (few) suites that build in `beforeAll`. That's a permanent CI/dev-loop tax on every future PR. Move the `npm run build` into a vitest `globalSetup` (runs exactly once, before any worker) and drop both the per-suite `beforeAll` build and the `fileParallelism` override. Since your ci.yml change already pre-builds, the globalSetup can even skip when `dist/` is present and newer than `src/`.

2. With (1) in place, the `execSync('npm run build')` inside `credentials.test.ts` goes away entirely — the child helper just consumes `dist/` like the other built-artifact suites.

Everything else (the lock semantics, the test moves to async/await, ci.yml build step) is good to go. One round on the test infra and this merges.

jangjos-128:
Hi @SahilRakhaiya05 — so sorry for the disruption here. 🙏 While we were publishing a new CLI release, a hiccup in our process unintentionally closed this PR. To be clear: **your contribution wasn't rejected**, and nothing on your end caused this.

PR #33, opened 2026-06-25
fix(rerun): preserve exit code and escalate auth errors in batch rerun

crypticsaiyan:
## Summary

`runTestRerun`'s `pollAccepted` hardcoded `exitCode: 1` on every `ApiError` during run polling. The batch aggregator detects auth failures via `rerunResults.find(r => r.error?.exitCode === 3)`. With every error stored as `exitCode: 1`, auth errors were never escalated and the batch always exited 1 instead of 3.

The sibling function `runTestRunAll` already had both fixes applied. `runTestRerun` missed them.

## The Fix

- **Exit Code Preservation:** `pollAccepted` now stores `exitCode: err.exitCode` instead of hardcoded `1`, so the aggregator sees the real error class from each polling failure.
- **Auth Escalation:** Added an `authErr` check before the generic exit-1 throw: any batch member with `exitCode: 3` (`AUTH_REQUIRED` / `AUTH_INVALID`) escalates the whole batch to exit 3 with an actionable message.
- **Idempotency Key Overflow (D4):** Initial chunk dispatch now truncates the base key so `base + :chunkN` never exceeds 256 chars, matching the identical guard already present in the deferred-retry path.

## Testing

Added 9 tests to `src/commands/test.rerun.spec.ts` across three describe blocks:

- `[fix-exitcode]`: `AUTH_REQUIRED` escalates to exit 3; `RATE_LIMITED` and `NOT_FOUND` remain exit 1 (no spurious escalation)
- `[fix-auth-escalation]`: auth failure exits 3; mixed pass+auth batch exits 3; non-auth failure exits 1
- `[fix-D4]`: short key passes through unchanged; 249-char key fits exactly at 256; 256-char key truncated to preserve suffix

## Files Changed

- `src/commands/test.ts`
- `src/commands/test.rerun.spec.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `test rerun` batch polling so authentication failures now return the correct exit code, even in mixed results.
  * Non-authentication errors continue to return a standard failure exit code.
  * Ensured chunked rerun requests keep derived keys within the allowed length limit, including edge cases near the boundary.

### DIFF
--- a/src/commands/test.rerun.spec.ts
+++ b/src/commands/test.rerun.spec.ts
@@ -4698,3 +4698,485 @@ describe('rerun --wait — dashboardUrl on terminal output', () => {
     );
   });
 });
+
+// ---------------------------------------------------------------------------
+// [fix-exitcode] pollAccepted preserves ApiError exit codes (not hardcoded 1)
+// ---------------------------------------------------------------------------
+
+describe('[fix-exitcode] polling error exit codes preserved in batch rerun results', () => {
+  it('AUTH_REQUIRED during polling → batch escalates to exitCode 3', async () => {
+    const creds = makeCreds();
+    const passRun = makeTerminalRun('run_pass_a1', 'passed');
+    const batchResp: BatchRerunResponse = {
+      accepted: [
+        { testId: 'test_1', runId: 'run_auth_fail', enqueuedAt: '2026-06-03T10:00:00.000Z' },
+        { testId: 'test_2', runId: 'run_pass_a1', enqueuedAt: '2026-06-03T10:00:00.000Z' },
+      ],
+      deferred: [],
+      conflicts: [],
+      closure: { byProject: [] },
+    };
+
+    const fetchImpl = makeFetch(url => {
+      if (url.includes('/tests/batch/rerun')) return { status: 202, body: batchResp };
+      if (url.includes('/runs/run_auth_fail')) return errorBody('AUTH_REQUIRED');
+      if (url.includes('/runs/run_pass_a1')) return { body: passRun };
+      return errorBody('NOT_FOUND');
+    });
+
+    const err = await runTestRerun(
+      {
+        testIds: ['test_1', 'test_2'],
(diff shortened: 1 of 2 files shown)


zeshi-du:
@coderabbitai review

crypticsaiyan:
Thanks @coderabbitai. I went through all four comments:

1 . **Malformed duplicate deps / trailing `.catch`** (`test.rerun.spec.ts` ~4727–4744): Not reproducible. Each block is a single valid `runTestRerun(...)` call followed by one `.catch(...)`. The suite compiles with `tsc --noEmit`, and all tests pass, so there isn't a syntactically incomplete statement here.

2 & 4. **`NOT_FOUND` polling should exit 4 instead of 1** (`spec` ~4791–4830 and `test.ts` ~6749–6761): This is intentional and matches the sibling `runTestRunAll` aggregation that this PR mirrors. That function folds every non-auth failure into the generic batch exit code `1` and special-cases only auth failures as exit code `3`, because invalid credentials affect the entire batch, whereas a `NOT_FOUND` during polling applies only to an individual member. The per-member status is still preserved in the JSON response. Returning exit code `4` here would make `rerun` behave differently from `runTestRunAll` and break that established contract, so the test asserting exit code `1` is correct.

3 . **Use a digest instead of prefix truncation for the chunk key** (`test.ts` ~6288–6296): Keeping prefix truncation intentionally because it matches the existing retry and deferred-retry key generation. This keeps all chunk-key paths consistent. The collision scenario described would require caller-supplied idempotency keys longer than 256 characters that also share roughly the first 250 characters, which is not a realistic input. Switching just this location to a digest would introduce inconsistency without providing a practical benefit.

crypticsaiyan:
Hey @zeshi-du @ruili-testsprite, kindly review this PR. Also, could you please assign issue #78 to me as it's referenced by this PR before merging? Thanks!

zeshi-du (approved):

PR #34, opened 2026-06-25
fix(test): reject empty code get --out and strip code-file BOM

SahilRakhaiya05:
## Summary

Two silent-failure fixes in the test command surface:

### 1. `test code get --out` with empty inline code

When the API returns an empty/null `code` body in text mode, the CLI printed a stderr hint and exited 0 - but if `--out` was set, `closeOutputFile()` still ran and left a **zero-byte file**. Agents and scripts checking file existence/size treated that as a successful download.

**Fix:** abort the file sink, unlink any artifact, and throw `VALIDATION_ERROR` (exit 5). Stderr-only text mode (no `--out`) behavior is unchanged.

### 2. UTF-8 BOM on `--code-file`

Plan/steps JSON reads already strip a leading BOM from PowerShell 5.1 `Set-Content -Encoding utf8` files. `readCodeFile()` did not - the invisible `U+FEFF` prefix was uploaded as part of the test source via `test code put` / `test create --code-file`.

**Fix:** apply existing `stripBom()` in `readCodeFile()`.

## Tests

- `runCodeGet`: empty inline code + `--out` ? exit 5, no file left behind; stderr hint preserved without `--out`
- `runCodePut`: BOM-prefixed code file uploads body without BOM

This Pr is  Solving Issue #117 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `test code get` so empty results now fail cleanly when saving to a file, without leaving behind empty or partial output files.
  * Preserved the existing “no code generated yet” hint when no output file is requested.
  * Uploads now handle files with a UTF-8 BOM correctly.

* **Chores**
  * Updated automated checks to run builds before tests and reduced flaky test behavior in the test suite.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,7 @@ jobs:
           cache: 'npm'
 
       - run: npm ci
+      - run: npm run build
       - run: npm test
         env:
           CI: true
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -1708,7 +1708,7 @@ describe('runCodePut', () => {
   it('strips a UTF-8 BOM from --code-file before uploading (Windows PowerShell 5.1 default)', async () => {
     const { credentialsPath } = makeCreds();
     const dir = mkdtempSync(join(tmpdir(), 'cli-p4-bom-'));
-    const codeFile = join(dir, 'updated.spec.ts');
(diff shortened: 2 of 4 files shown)


zeshi-du:
@coderabbitai review

zeshi-du (approved):

PR #39, opened 2026-06-25
fix(test): reject whitespace-only --name in test update (parity with test create)

### LINKED ISSUE #1: Video asset

lxcario:
## What

`test update --name "   "` (whitespace-only) is accepted and would send a blank name to the backend, while the sibling `test create --name "   "` correctly rejects it with `VALIDATION_ERROR` (exit 5).

## Why this matters

A whitespace-only name produces a junk test record that is effectively invisible in `test list` output and the dashboard. The `test create` path already prevents this via `requireString` (dogfood P1 fix #1); `test update` missed the same guard and only checks the upper length bound.

## Reproduction

```bash
# test update -- whitespace name ACCEPTED (exit 0), would send "   " to backend:
$ testsprite test update test_abc --name "   " --dry-run --output json
{ "testId": "test_dryrun_update_2026", "updatedFields": ["name"], ... }   # exit 0

# control: test create -- whitespace name correctly REJECTED (exit 5):
$ testsprite test create --project proj_1 --type frontend --name "   " --code-file x.ts --dry-run --output json
{ "error": { "code": "VALIDATION_ERROR", "nextAction": "Flag `--name` is invalid: is required." } }   # exit 5
```

## Root cause

`src/commands/test.ts`, `runUpdate` (line ~1212). The function validates name length:

```ts
if (opts.name !== undefined && opts.name.length > 200) {
  throw localValidationError('name', 'must be at most 200 characters');
}
```

But has no lower-bound / whitespace check. A whitespace-only string passes `opts.name !== undefined` (it's a non-empty string), the length check (3 < 200), and flows into the PATCH body.

## Fix

Add a whitespace rejection before the length check, mirroring the existing `requireNonEmpty` / `requireString` pattern used elsewhere in the same file:

```ts
if (opts.name !== undefined && opts.name.trim().length === 0) {
  throw localValidationError('name', 'must be a non-empty string (whitespace-only is not allowed)');
}
```

## Tests

Added 1 regression test to `src/commands/test.test.ts` (in the `runUpdate` describe block) asserting `name: '   '` rejects with `VALIDATION_ERROR` / field `name` and makes no network call. Fails on `main` before this fix; passes after.

- After (this branch): `Tests 16 failed | 1360 passed | 72 skipped` (+1 new test)

The 16 failures are pre-existing and environment-specific (Windows path/line-ending), unrelated -- see #4.

## Verification

- `npm run typecheck`: pass
- `npm run lint`: pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for the `--name` option so whitespace-only values are now rejected before any update request is sent.
  * Invalid input now returns a clear validation error associated with the `name` field.
* **Tests**
  * Added coverage to ensure whitespace-only `--name` inputs fail validation and do not trigger any network call.

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -5777,6 +5777,31 @@ describe('runUpdate', () => {
     expect(called).toBe(0);
   });
 
+  it('rejects a whitespace-only --name before sending (parity with test create)', async () => {
+    const { credentialsPath } = makeCreds();
+    let called = 0;
+    const fetchImpl = makeFetch(() => {
+      called += 1;
+      return { body: SAMPLE_RESPONSE };
+    });
+    await expect(
+      runUpdate(
+        {
+          profile: 'default',
+          output: 'json',
+          debug: false,
+          testId: 'test_alpha',
+          name: '   ',
+        },
+        { credentialsPath, fetchImpl, stdout: () => undefined },
+      ),
+    ).rejects.toMatchObject({
+      code: 'VALIDATION_ERROR',
+      details: expect.objectContaining({ field: 'name' }),
+    });
+    expect(called).toBe(0);
+  });
+
   it('renders text mode with one line per updated field', async () => {
     const { credentialsPath } = makeCreds();
     const fetchImpl = makeFetch(() => ({ body: SAMPLE_RESPONSE }));
--- a/src/commands/test.ts
+++ b/src/commands/test.ts
@@ -1244,7 +1244,10 @@ export async function runUpdate(
   requireNonEmpty('test-id', opts.testId);
   // P1-3: client-side length checks matching server limits.
   if (opts.name !== undefined && opts.name.trim().length === 0) {
-    throw localValidationError('name', 'must be a non-empty string (whitespace-only is not allowed)');
+    throw localValidationError(
+      'name',
+      'must be a non-empty string (whitespace-only is not allowed)',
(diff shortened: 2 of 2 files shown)


zeshi-du:
@coderabbitai review

zeshi-du:
Thanks @lxcario! 🙏 CI is green except one thing — Prettier flagged an unformatted file:

```
[warn] src/commands/test.ts
Code style issues found in the above file. Run Prettier with --write to fix.
```

One command on your branch fixes it:

```
npm run format && git commit -am "style: run prettier"
```

Typecheck, unit, e2e, coverage and CodeRabbit are all green — this is the only thing standing. Push that and it's good to merge. 🚀

lxcario:
Thanks @zeshi-du! Ran the formatter on src/commands/test.ts -- it's now Prettier-clean and pushed. Typecheck, unit, e2e and coverage were already green, so this should clear the last check. Ready for another look whenever you have a moment.

lxcario:
Rebased onto current main; all 8 checks are green. Whitespace-only --name in test update is rejected before any network call, mirroring test create, with a regression test asserting the validation error and that fetchImpl is not called. Ready when you are.

zeshi-du (approved):

PR #41, opened 2026-06-26
fix(help): include request-timeout in global flag guidance

JerryNee:
## What

`--request-timeout` is a global flag, but it was missing from two pieces of global-flag guidance:

- the helper that rephrases misplaced global flags after subcommands
- the shared `GLOBAL_OPTS_HINT` footer appended to subcommand help text

## Why

When a user places a known global flag after a subcommand, the CLI tries to turn Commander’s generic `unknown option` message into an actionable hint. `--request-timeout` should get the same treatment as `--output`, `--profile`, and `--endpoint-url`.

## Changes

- Add `request-timeout` to the global flag arity table as a value flag.
- Include `--request-timeout` in the shared global options help footer.
- Add regression tests for the rephrased misplaced-flag message.
- Refresh help snapshots for the updated footer.

## Verification

- `npx vitest run src/lib/render-error.test.ts test/help.snapshot.test.ts -u`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm test`

### DIFF
--- a/src/lib/output.ts
+++ b/src/lib/output.ts
@@ -3,10 +3,10 @@ export type OutputMode = 'json' | 'text';
 /**
  * Help-text footer pointing at the global options surface so users
  * looking at any subcommand `--help` don't miss `--dry-run`, `--output`,
- * `--profile`, `--endpoint-url`, `--debug`.
+ * `--profile`, `--endpoint-url`, `--request-timeout`, `--debug`.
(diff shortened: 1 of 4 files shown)


zeshi-du:
@coderabbitai review

zeshi-du (approved):

PR #47, opened 2026-06-26
fix(test): validate artifact run id default path

### LINKED ISSUE #45: fix(test): prevent path-like run IDs escaping artifact default output dir
## Summary

`testsprite test artifact get <run-id>` uses `<run-id>` directly as the final path segment for its default output directory:

```text
./.testsprite/runs/<run-id>/
```

Because the value is not validated as a path-safe segment, a path-like `runId` such as `../../outside` can make the default bundle directory resolve outside the documented `.testsprite/runs/` root when `--out` is omitted.

This is a local filesystem safety issue in the CLI default path construction. It is separate from explicit `--out` handling: users can already choose any custom output location with `--out`; the default path should remain constrained to the CLI-owned artifact tree.

## Why this matters

The command is agent-facing and is often copied from CLI output after a failed run:

```bash
testsprite test artifact get <run-id>
```

A caller reasonably expects the default destination to be under:

```text
./.testsprite/runs/<run-id>/
```

But if `<run-id>` contains path separators or traversal segments, Node path joining treats it as path structure rather than an opaque identifier. That can cause the bundle writer to create or overwrite files outside the documented artifact root.

Even if production run IDs are normally generated by TestSprite, the CLI should fail closed here because:

- `runId` is a positional user input, not a trusted local path.
- `artifact get` writes multiple files and directories.
- `--out` already exists for intentional custom paths.
- The documented default path implies containment under `.testsprite/runs/`.

## Reproduction

From a clean checkout, call `runArtifactGet` without `--out` and with a path-like run id:

```ts
await runArtifactGet(
  {
    profile: 'default',
    output: 'json',
    debug: false,
    runId: '../../outside',
    failedOnly: false,
  },
  { fetchImpl, stdout: () => {} },
);
```

Current behavior: the default directory is constructed from the raw `runId`, equivalent to:

```ts
join(process.cwd(), '.testsprite', 'runs', '../../outside')
```

which normalizes outside `./.testsprite/runs/`.

A CLI-level reproduction is the same shape:

```bash
testsprite test artifact get ../../outside --dry-run --output json
```

The dry-run path is especially useful as a contract check because it reports the default `out` value without network or disk writes.

## Expected behavior

When `--out` is omitted, `<run-id>` should be treated as an opaque identifier and must be a single path-safe segment. Values containing path separators, traversal segments, or NUL bytes should fail before auth/network/disk work with a typed validation error:

```json
{
  "error": {
    "code": "VALIDATION_ERROR",
    "details": {
      "field": "run-id"
    }
  }
}
```

If a caller intentionally wants a custom directory, they should use the existing explicit path mechanism:

```bash
testsprite test artifact get <run-id> --out ./some/custom/dir
```

## Suggested fix

Add a small helper for the default artifact directory path:

```ts
function resolveDefaultArtifactDir(runId: string, cwd = process.cwd()): string {
  requireNonEmpty('run-id', runId);
  if (runId === '.' || runId === '..' || runId.includes('/') || runId.includes('\\')) {
    throw localValidationError(
      'run-id',
      'must be a single path-safe segment for the default output directory; pass --out <dir> to choose a custom path',
    );
  }
  if (runId.includes('\0')) {
    throw localValidationError('run-id', 'must not contain NUL bytes');
  }
  return join(cwd, '.testsprite', 'runs', runId);
}
```

Then use it in `runArtifactGet` only for the implicit default path:

```ts
const resolvedDir =
  opts.out !== undefined
    ? resolveBundleDir(opts.out)
    : resolveDefaultArtifactDir(runId);
```

It is also worth resolving/validating the output path before constructing the HTTP client so invalid local input fails without requiring credentials or making a request.

## Acceptance criteria

- `test artifact get ../../outside` without `--out` exits with `VALIDATION_ERROR` before auth/fetch.
- Normal run IDs still use the documented default directory: `./.testsprite/runs/<run-id>/`.
- Explicit `--out <dir>` behavior remains unchanged.
- Add regression coverage for both unsafe path-like run IDs and normal run IDs.

Lexiie:
## Summary

Fixes #45.

`test artifact get <run-id>` used the raw positional `runId` as the final segment of the default output directory:

```text
./.testsprite/runs/<run-id>/
```

That made the implicit destination path interpret path-like run IDs as filesystem structure. For example, a value like `../../outside` would normalize outside the documented `.testsprite/runs/` artifact root when `--out` was omitted.

This PR treats `runId` as an opaque identifier for the implicit default path. If callers want a custom filesystem location, the existing explicit `--out <dir>` path remains the supported mechanism.

## Changes

- Add `resolveDefaultArtifactDir(runId, cwd)` for the implicit artifact output directory.
- Reject path-like run IDs for the default path:
  - `.` / `..`
  - `/`
  - `\`
  - NUL bytes
- Keep normal run IDs on the documented path: `./.testsprite/runs/<run-id>/`.
- Move HTTP client construction until after local path validation and dry-run handling, so invalid local input fails before credentials, network, or disk work.
- Add regression coverage for unsafe path-like run IDs and the normal default path.

## Why

`test artifact get` writes a multi-file failure bundle. The documented default path implies containment under `.testsprite/runs/`, while `--out` is already available for intentional custom paths. Failing closed on path-like positional IDs keeps the default artifact tree predictable and avoids surprising local writes.

## Verification

- `npx vitest run src/commands/test.artifact.spec.ts`
- `npm run typecheck`

### DIFF
--- a/src/commands/test.artifact.spec.ts
+++ b/src/commands/test.artifact.spec.ts
@@ -20,6 +20,7 @@ import type { CliFailureContext, CliLatestResult, CliTestStep } from './test.js'
 import {
   assertOutDirParentExists,
   createTestArtifactCommand,
+  resolveDefaultArtifactDir,
   runArtifactGet,
   runFailureGet,
 } from './test.js';
@@ -317,6 +318,31 @@ describe('runArtifactGet', () => {
     } finally {
       cwdSpy.mockRestore();
     }
+  });
+
+  it('rejects path-like runId before auth or fetch when default --out is used', async () => {
+    const fetchImpl = vi.fn<typeof globalThis.fetch>();
+
+    await expect(
+      runArtifactGet(
+        {
+          profile: 'default',
+          output: 'json',
+          debug: false,
+          runId: '../../outside',
+          failedOnly: false,
+        },
+        { fetchImpl, stdout: () => {} },
+      ),
+    ).rejects.toMatchObject({ code: 'VALIDATION_ERROR', exitCode: 5 });
+
+    expect(fetchImpl).not.toHaveBeenCalled();
+  });
+
+  it('keeps the documented default directory for path-safe runIds', () => {
+    expect(resolveDefaultArtifactDir(SAMPLE_RUN_ID, '/repo')).toBe(
+      join('/repo', '.testsprite', 'runs', SAMPLE_RUN_ID),
+    );
   });
 
   // ---- --failed-only passed through to writeBundle ----
--- a/src/commands/test.ts
+++ b/src/commands/test.ts
@@ -6703,6 +6703,20 @@ export interface ArtifactGetResult {
   bundle?: WriteBundleResult;
 }
 
+export function resolveDefaultArtifactDir(runId: string, cwd: string = process.cwd()): string {
+  requireNonEmpty('run-id', runId);
+  if (runId === '.' || runId === '..' || runId.includes('/') || runId.includes('\\')) {
+    throw localValidationError(
+      'run-id',
+      'must be a single path-safe segment for the default output directory; pass --out <dir> to choose a custom path',
+    );
+  }
+  if (runId.includes('\0')) {
+    throw localValidationError('run-id', 'must not contain NUL bytes');
+  }
+  return join(cwd, '.testsprite', 'runs', runId);
+}
+
 /**
  * Validate that the parent directory of `resolvedDir` exists and is a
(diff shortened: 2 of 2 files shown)


zeshi-du:
@coderabbitai review

zeshi-du:
Thanks for the security-minded fix, @Lexiie! 🙏

#71 guards the artifact `runId` path the same way and additionally handles the Windows trimmed-dot bypass (`. ` / `.. `), so we're consolidating on #71 and closing this as a duplicate in its favor.

Genuinely appreciate the contribution — please keep them coming! 🚀

PR #14, opened 2026-06-24
fix(cli): validate --output uniformly across all command groups

Davidson3556:
Fixes #13

#### Describe the changes you have made in this PR -

The global `--output` flag was validated inconsistently across command groups. `test` and `project` rejected an unrecognised value with a typed `VALIDATION_ERROR` (exit 5), but `auth`, `usage`, `agent`, and `init` resolved it with `globals.output ?? 'text'` — silently coercing anything that isn't `json`/`text` to **text mode** and exiting `0`.

For the CLI's primary consumer (a coding agent), that's a footgun: a request for `--output json` with a typo (`--output josn`) returned a human-readable text payload, so the agent's downstream `JSON.parse` failed with no signal as to why. The two commands that *did* validate also used different wording.

Changes:
- Add a single `resolveOutputMode(raw)` helper in `src/lib/output.ts`. It returns `'text'` for an omitted flag and throws `localValidationError('output', 'must be one of: json, text', ['json', 'text'])` for any other value.
- Route every command group's `resolveCommonOptions` (`auth`, `usage`, `agent`, `init`, `project`, `test`) through it — closing the validation gap on the four unvalidated groups and unifying the error wording on the canonical `test`-style message.
- Add unit tests for `resolveOutputMode` and an end-to-end subprocess regression for two previously-unvalidated groups (`agent list`, `auth status`).

Note: when `--output` itself is the invalid value, the error envelope is rendered in text mode (the requested mode is unusable) — unchanged from the existing `test`/`project` behaviour.

### Demo/Screenshot for feature changes and bug fixes -

Before (main) — invalid value silently coerced to text, exit 0:

Screenshot 2026-06-24 at 00 15 33

After (this branch) — rejected with an actionable message; valid values unaffected:

image

Tests:
Screenshot 2026-06-23 at 23 53 07

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

[...]

PR #50, opened 2026-06-28
fix(agent): reject invalid output modes

JerryNee:
## Summary

- validate `testsprite agent ... --output <mode>` against the supported `json|text` modes
- return the existing local `VALIDATION_ERROR` shape instead of silently treating unsupported modes as text
- add command-wiring coverage for `agent list`

## Before

`testsprite agent list --output yaml` exited 0 and printed text output.

## After

`testsprite agent list --output yaml` exits 5 with:

```text
Error: Invalid request.
Flag `--output` is invalid: must be one of: json, text.
```

## Verification

- `npm test -- src/commands/agent.test.ts`
- `npm run lint`
- `npm run typecheck`
- `npm test`
- `npm run build`
- `node dist/index.js agent list --output yaml`

### DIFF
--- a/src/commands/agent.test.ts
+++ b/src/commands/agent.test.ts
@@ -854,6 +854,28 @@ describe('createAgentCommand wiring', () => {
     expect(out).toContain('claude');
     expect(out).toContain('antigravity');
   });
+
+  it('agent list rejects unsupported global --output modes', async () => {
+    const { deps } = makeCapture();
+
+    const command = createAgentCommand({ cwd: CWD, ...deps });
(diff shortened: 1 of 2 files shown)


zeshi-du:
@coderabbitai review

zeshi-du:
Thanks @JerryNee! 🙏

#14 introduces a shared `resolveOutputMode()` helper that applies this same `--output` validation across every command group in one place (including a few the per-command PRs didn't cover), so we're going with #14 as the single source of truth. Closing this as a duplicate in its favor.

Great catch on the gap — hope to see more PRs from you! 🚀

PR #52, opened 2026-06-28
fix(auth): honor request timeout during configure

JerryNee:
## Summary

- pass `requestTimeoutMs` into the pre-write `/me` validation client used by `auth configure` / `setup`
- ensure a configured short request timeout also bounds the key-validation ping before credentials are written
- add regression coverage that proves the validation fetch receives an abort signal and credentials are not persisted on timeout

## Why

`runConfigure` builds a one-off `HttpClient` so it can validate the candidate key before writing credentials. That path bypassed `makeHttpClient`, so the global `--request-timeout` value was parsed but not applied to the validation ping.

## Verification

- `npm test -- src/commands/auth.test.ts`
- `npm run lint`
- `npm run typecheck`
- `npm test`
- `npm run build`

### DIFF
--- a/src/commands/auth.test.ts
+++ b/src/commands/auth.test.ts
@@ -110,6 +110,50 @@ describe('runConfigure', () => {
     );
   });
 
+  it('uses requestTimeoutMs for the pre-write key validation ping', async () => {
+    const { deps } = makeCapture();
+    let sawAbort = false;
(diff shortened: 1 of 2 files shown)


zeshi-du:
@coderabbitai review

zeshi-du (approved):

PR #53, opened 2026-06-28
fix(timeout): reject invalid request-timeout flags

JerryNee:
## Summary
- centralize parsing for the global `--request-timeout <seconds>` flag
- reject explicit invalid values such as `abc`, `0`, `1.5`, and `601` with a local VALIDATION_ERROR instead of silently using/rounding/clamping another timeout
- keep env/direct factory timeout resolution behavior unchanged while wiring auth/project/test/usage/setup through the shared flag parser

## Repro
Before this change, `testsprite project list --dry-run --request-timeout abc` exited 0 and used the default timeout. After this change it exits 5 with a machine-readable VALIDATION_ERROR naming `request-timeout`.

## Verification
- `npm test -- src/lib/client-factory.test.ts`
- `npm test -- test/cli.subprocess.test.ts`
- `npm run lint`
- `npm run typecheck`
- `npm test`
- `npm run build`
- `node dist/index.js --output json project list --dry-run --request-timeout abc`

### DIFF
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -2,6 +2,7 @@ import { Command } from 'commander';
 import {
   emitDryRunBanner,
   makeHttpClient,
+  parseRequestTimeoutFlag,
   type CommonOptions as FactoryCommonOptions,
 } from '../lib/client-factory.js';
 import type { ErrorCode } from '../lib/errors.js';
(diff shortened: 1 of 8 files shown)


zeshi-du:
@coderabbitai review

zeshi-du:
Thanks for tackling this, @JerryNee! 🙏

We're consolidating `--request-timeout` validation on #17 — it lands the same fix, de-duplicates the parser across all commands, and preserves the documented fractional-seconds/clamp behavior. So I'm closing this as a duplicate in favor of #17.

One idea worth pursuing separately: if you'd like stricter integer-only validation, that'd make a great standalone PR alongside a `DOCUMENTATION.md` update. Really appreciate the contribution — please keep them coming! 🚀

PR #54, opened 2026-06-28
fix(project): avoid password file reads in dry-run update

JerryNee:
## Summary
- keep `project update --dry-run` from reading `--password-file` before the dry-run preview
- compute `updatedFields` from supplied flags, then resolve password file contents only on the real PATCH path
- add unit and subprocess regression coverage for a missing password file under dry-run

## Repro
Before this change, `testsprite project update proj_anything --password-file /tmp/missing --dry-run --output json` exited 1 with a raw ENOENT, even though dry-run is documented to skip filesystem writes/reads. After this change it exits 0 and previews `updatedFields: ["password"]`.

## Verification
- `npm test -- src/commands/project.test.ts`
- `npm test -- test/cli.subprocess.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- `npm run build`
- `node dist/index.js project update proj_anything --password-file /tmp/definitely-not-here-testsprite --dry-run --output json`

### DIFF
--- a/src/commands/project.test.ts
+++ b/src/commands/project.test.ts
@@ -668,6 +668,33 @@ describe('runUpdate', () => {
     expect(result.updatedFields).toContain('name');
   });
 
+  it('P7 — dry-run with --password-file does not read the filesystem', async () => {
+    const { credentialsPath } = makeCreds();
+    const fetchImpl = vi.fn(async () => {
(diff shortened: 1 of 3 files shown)


zeshi-du:
@coderabbitai review

zeshi-du (approved):

PR #55, opened 2026-06-28
fix(history): reject fractional page sizes

JerryNee:
## Summary
- make `test result --history --page-size` reject fractional values before building the request
- keep the existing 1-100 bounds check, but add an integer check for values such as `1.5`
- add regression coverage that ensures no request is made for a fractional history page size

## Repro
Before this change, `testsprite test result test_abc --history --dry-run --page-size 1.5 --output json` exited 0 and passed the fractional page size through. After this change it exits 5 with a local VALIDATION_ERROR naming `page-size`.

## Verification
- `npm test -- src/commands/test.result.history.spec.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- `npm run build`
- `node dist/index.js --output json test result test_abc --history --dry-run --page-size 1.5`

### DIFF
--- a/src/commands/test.result.history.spec.ts
+++ b/src/commands/test.result.history.spec.ts
@@ -533,6 +533,32 @@ describe('runResultHistory — pagination', () => {
 
     expect(capturedUrl).toContain('pageSize=5');
   });
+
+  it('rejects fractional --page-size before making a request', async () => {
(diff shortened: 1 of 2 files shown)


zeshi-du:
@coderabbitai review

zeshi-du (approved):

PR #48, opened 2026-06-27
fix(paginate): bound cursor loops without dropping empty pages

aldorizona10-glitch:
## Summary

Fixes the pagination infinite-loop risk in `paginate()` without dropping legitimate empty cursor pages.

## Problem

The original bug report covered a pagination loop where an endpoint can keep returning no items with a non-null cursor. A naive fix would stop on `items.length === 0`, but maintainers pointed out that filtered cursor stores can legitimately return an empty page with a cursor while later pages still contain matching results.

## Fix

`paginate()` now keeps consuming empty pages, but adds two progress bounds:

- detect a repeated non-null `nextToken` and reject with a typed `UNAVAILABLE` error
- cap auto-pagination at `MAX_AUTO_PAGES = 1000` and reject if the cap is exceeded

This prevents infinite loops while preserving valid filtered pagination behavior.

## Changes

- `src/lib/pagination.ts`
  - adds `MAX_AUTO_PAGES`
  - tracks seen non-null cursors
  - throws typed pagination safety errors for `repeated_next_token` and `max_pages_exceeded`
- `src/lib/pagination.test.ts`
  - verifies empty cursor pages can lead to later data
  - verifies repeated cursor detection
  - verifies max-page safety cap behavior
- removes the duplicate `test/pagination.test.ts` file and keeps coverage in the existing pagination test module

## Testing

- `npm test -- src/lib/pagination.test.ts`
- `npm run format:check`
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `npm test` — 42 files / 1486 tests passed

Fixes #35

Signed-off-by: Aldo Rizona <aldorizona10@gmail.com>

PR #40, opened 2026-06-26
fix(pagination): reject fractional pagination flags

JerryNee:
## What

`--page-size` and `--max-items` are documented as integer pagination values, but shared pagination validation accepted fractional numbers like `1.5` and `2.5`.

## Why

Fractional pagination parameters are invalid per CLI docs/error wording and can produce surprising API requests. Local validation should reject them before credentials or network work.

## Changes

- Require `Number.isInteger` for `pageSize` and `maxItems`.
- Add regression tests for fractional `pageSize` and `maxItems`.

## Verification

- `npx vitest run src/lib/pagination.test.ts src/commands/project.test.ts src/commands/test.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm test`
- `npm run build`

PR #61, opened 2026-06-30
fix: bound pagination and guard password file reads

### LINKED ISSUE #58: Unbounded Pagination and Arbitrary File Read via --password-file
## Summary

- performs `statSync()` first
- enforces 350 KB size limit

---

## Affected Code

```typescript
// src/lib/pagination.ts:95
while (true) {
  cursor = page.nextToken;
}
```

## Impact when exploited  

merlinsantiago982-cmd:
## Summary
- stop auto-pagination when the server repeats a cursor or exceeds a hard page budget
- guard `--password-file` reads with stat-first validation, regular-file checks, and a 64 KiB size cap
- add tests for cursor cycles, runaway pagination, valid password files, directory rejection, and oversized password files

## Why
A malicious or broken server could keep returning non-null cursors indefinitely, causing the CLI to loop and grow memory. `--password-file` also read arbitrary paths without checking the target type or size before sending the content as the project password.

## Testing
- npm test -- src/lib/pagination.test.ts src/commands/project.test.ts
- npm run typecheck
- npm run format:check
- npm run lint
- npm run build

Fixes #58

## Summary by CodeRabbit

* **New Features**
  * Added safer handling for password files when creating or updating projects, including file validation and a size limit.
  * Added a limit for automatic pagination to prevent runaway requests.

* **Bug Fixes**
  * Rejects invalid password-file paths, oversized password files, and repeated pagination cursors before making network calls.
  * Prevents infinite or excessive auto-pagination by stopping at a configured maximum page count.


zeshi-du (dismissed):

zeshi-du:
Thanks — the pagination safety cap + loop detection and the guarded password-file reader (regular-file check, 64 KB cap, clear ENOENT error) are all improvements we want. This PR now conflicts with today's merge wave: #54 restructured the password-resolution block in `runUpdate`, #40 landed fractional-flag validation in `pagination.ts`, and #14/#17 touched the import blocks. Please rebase on latest `main` — the substance of your change is pre-approved, so a clean rebase should merge quickly. Note #27/#106 touch the same lines; your guarded reader supersedes #27's try/catch, so don't be surprised if the region looks different again after they land.

zeshi-du:
Thanks for the patience here — an update on scope. The pagination-bounding half of this PR just landed via #48 (same design you proposed: page safety cap + repeated-cursor detection; that PR rebased first while this one has been conflicting since 2026-07-02). The **password-file read guard** (stat + regular-file check + size cap) is still very much wanted — it's the remaining open half of issue #58. If you rebase this PR down to just the password-file guard (drop the `pagination.ts` changes), it merges quickly. 🙏

jangjos-128:
Hi @merlinsantiago982-cmd — so sorry for the disruption here. 🙏 While we were publishing a new CLI release, a hiccup in our process unintentionally closed this PR. To be clear: **your contribution wasn't rejected**, and nothing on your end caused this.

PR #71, opened 2026-06-30
fix(test): guard default artifact run id path

### LINKED ISSUE #45: fix(test): prevent path-like run IDs escaping artifact default output dir

merlinsantiago982-cmd:
## Summary
- add a default artifact directory resolver that treats `run-id` as one path-safe segment
- reject path-like or NUL-containing run IDs before auth/fetch when `--out` is omitted
- keep explicit `--out` behavior unchanged

## Testing
- npm test -- src/commands/test.artifact-path.spec.ts
- npm run typecheck
- npm run format:check
- npm run lint
- npm run build

Fixes #45

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved default artifact download location to reliably save results under the expected runs directory for the given run ID.
  * Added stricter run ID validation, preventing unsafe values from being used unless a specific output directory is provided.
* **Tests**
  * Added regression tests covering default path containment, early rejection of unsafe run IDs, and behavior when an explicit output directory is set.

### DIFF
--- a/src/commands/test.artifact-path.spec.ts
+++ b/src/commands/test.artifact-path.spec.ts
@@ -1,0 +1,56 @@
+/**
+ * Regression coverage for default artifact output path containment.
+ */
+
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { describe, expect, it } from 'vitest';
+import { resolveDefaultArtifactDir, runArtifactGet } from './test.js';
+
+describe('resolveDefaultArtifactDir', () => {
+  it('keeps a normal run id under .testsprite/runs', () => {
+    const cwd = join(tmpdir(), 'artifact-path');
+    expect(resolveDefaultArtifactDir('run_abc123', cwd)).toBe(
+      join(cwd, '.testsprite', 'runs', 'run_abc123'),
+    );
+  });
+
+  it.each(['.', '..', '../outside', '..\\outside', 'nested/run', 'nested\\run', 'bad\0id'])(
+    'rejects path-like run id %j',
+    runId => {
+      expect(() => resolveDefaultArtifactDir(runId, '/tmp/work')).toThrowError(
+        expect.objectContaining({
+          code: 'VALIDATION_ERROR',
+          details: expect.objectContaining({ field: 'run-id' }),
+        }),
+      );
+    },
+  );
+});
+
+describe('runArtifactGet default output path validation', () => {
+  it('rejects unsafe default run ids before auth or fetch work', async () => {
+    let fetchCalls = 0;
+    const fetchImpl = (async () => {
+      fetchCalls++;
+      return new Response('{}', { status: 200 });
+    }) as typeof globalThis.fetch;
+
+    await expect(
+      runArtifactGet(
+        {
+          profile: 'default',
+          output: 'json',
+          debug: false,
+          runId: '../../outside',
+          failedOnly: false,
+        },
+        { fetchImpl, stdout: () => {} },
+      ),
+    ).rejects.toMatchObject({
+      code: 'VALIDATION_ERROR',
+      details: expect.objectContaining({ field: 'run-id' }),
+    });
+    expect(fetchCalls).toBe(0);
+  });
+});
--- a/src/commands/test.artifact-path.spec.ts
+++ b/src/commands/test.artifact-path.spec.ts
@@ -3,9 +3,60 @@
  */
 
 import { tmpdir } from 'node:os';
+import { mkdirSync, mkdtempSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { describe, expect, it } from 'vitest';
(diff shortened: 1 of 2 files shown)


zeshi-du:
Thanks @merlinsantiago982-cmd! 🙏 Just one thing keeping CI red — a typecheck error:

```
src/commands/test.artifact-path.spec.ts(42,7): error TS2322:
Type '{ passed: number; failed: number; skipped: number; }' is not assignable to type 'string'.
```

Line 42 assigns a summary object where a `string` is expected — looks like a fixture/mock shape drifted from the field's type. Fix that assignment (or the field type) so `npm run typecheck` passes. Everything else + CodeRabbit are green. 🙏

zeshi-du:
Closing: the typecheck error flagged on 2026-07-01 (fixture `summary` shape) never got a push, and the same fix landed today via #172 with the identical structure (`resolveDefaultArtifactDir` + client-creation deferral) plus Windows trailing-dot/space normalization (`'...'`-style ids are collapsed by the Windows filesystem, so `trimEnd()` alone let them through). You proposed this guard first — that's on the record here — and issue #45 is resolved by the merge. The rescope suggestion on #61 (password-file guard) still stands if you'd like to land that piece. 🙏

PR #118, opened 2026-07-01
fix(prompt): preserve buffered input between prompts

cmdr-chara:
## Summary

Fixes prompt input handling so answers already buffered after the first newline are preserved for later prompts on the same input stream.

## Why

When input arrives as one buffered chunk, such as `first\nsecond\n`, the prompt reader returned `first` and discarded the remaining text. A second prompt on the same stream would then wait for input that had already been read.

## Changes

- Preserve unconsumed text after CR/LF terminators per input stream.
- Treat CRLF as one newline so sequential prompts do not receive an empty answer.
- Resolve buffered tail input when the stream has already ended.
- Add regression tests for sequential text prompts, EOF tail input, CRLF input, and secret prompts.

## Testing

- `npx vitest run src/lib/prompt.test.ts --reporter=dot`
- `npm run typecheck`
- `npm run build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved prompt input handling so sequential prompts on the same stream now correctly preserve and reuse buffered answers.
  * Fixed cases where the next prompt could miss leftover input at end-of-file or when using different line endings.
  * Ensured secret prompts continue to accept buffered input across multiple prompts without echoing answers to the screen.

### DIFF
--- a/src/lib/prompt.test.ts
+++ b/src/lib/prompt.test.ts
@@ -39,6 +39,31 @@ describe('promptText', () => {
     expect(await promptText('? ', { input, output })).toBe('ok');
   });
 
+  it('preserves buffered answers for sequential prompts on the same stream', async () => {
+    const input = Readable.from(['first\nsecond\nthird\n']);
+    const output = new CaptureStream();
+
+    await expect(promptText('One: ', { input, output })).resolves.toBe('first');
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):

PR #129, opened 2026-07-02
fix(agent): apply symlink fail-close guard to own-file --dry-run installs

ahndohun:
## What does this PR do?

`agent install --dry-run` for the own-file targets (`claude`, `cursor`, `cline`, `antigravity`) reports a would-be write without running the symlink fail-close guard, so it can claim success for an install the real run would refuse with exit 5 (planted symlink at the target file or at a parent path component). The codex managed-section branch already applies exactly this guard in dry-run (see the `[P2]` comment in `runInstall`): dry-run is supposed to be a truthful preview of the real install.

This PR runs the same `inspectTargetPath` check in the own-file dry-run branch, with the same error and exit code as the real path, and adds regression tests for both the symlinked-target and symlinked-parent cases.

Failure mode before this fix: `testsprite agent install --target claude --dry-run` on a repo where `.claude/skills/testsprite-verify/SKILL.md` (or `.claude` itself) is a symlink prints a dry-run success, but the actual install then fails with exit 5 — the preview lies about what the command will do.

## Related issue

None — small parity fix, opened directly as a PR per CONTRIBUTING.md ("Small fixes and improvements: just open a pull request").

## Type of change

- [ ] Documentation only
- [ ] Build / CI / chore

## Checklist

- [x] Commits follow Conventional Commits (`fix(agent): ...`).
- [x] `npm test` passes (1570/1570) and coverage stays at or above the 80% gate.
- [x] User-facing changes are reflected in `README.md` / `DOCUMENTATION.md` where relevant (no doc change needed — dry-run now simply matches the documented real behavior).

## Notes for reviewers

- Red-first: both new tests fail on `main` (dry-run reports success through the planted symlink) and pass with the fix.
- The guard reuses `inspectTargetPath` and the exact non-regular-file error of the real path, so refusal messages are identical between preview and real install.
- Track B / CLI Improvement Bonus: Discord `ahndohun`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dry-run installs now use the same safety checks as real installs, so they correctly fail when the target path is a symlink or otherwise not a regular file.
  * Dry-run behavior now matches actual install behavior by stopping with the same error and avoiding any writes.

### DIFF
--- a/src/commands/agent.test.ts
+++ b/src/commands/agent.test.ts
@@ -1311,6 +1311,50 @@ describe('runInstall — symlink safety', () => {
     expect(writeCalls.length).toBe(0); // never wrote a .bak nor through the link
   });
 
+  it('dry-run: refuses (exit 5) when the target file is a symlink (parity with real install)', async () => {
+    const { fs: agentFs, writeCalls, seedSymlink } = makeMemFs();
+    // Same planted SKILL.md symlink as the real-install case above: dry-run
+    // must report the same refusal the real install would, not a success.
+    seedSymlink(path.resolve(CWD, TARGETS.claude.path));
+    const { deps } = makeCapture();
+
+    let thrown: unknown;
+    try {
+      await runInstall(
+        { ...BASE_OPTS, target: ['claude'], force: false, dryRun: true },
+        { cwd: CWD, fs: agentFs, ...deps },
+      );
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):

PR #131, opened 2026-07-02
fix(credentials): strip CR/LF from values to prevent INI injection

lxcario:
## What

When the credentials file is serialized, a field value containing a newline (CR/LF) is written across multiple lines. Because the file is INI-style (key = value), an embedded newline lets an attacker inject arbitrary key/value pairs -- or a whole new section header -- into the credentials file.

## Why it matters

A valid API key or endpoint URL never contains a newline. But a value that reaches serialization from a compromised environment variable or a MITM'd backend response could smuggle a newline and override the stored api_key that is read back on the next invocation. This is a write-path hardening for a security-sensitive file.

## Fix

In serializeCredentials, strip CR/LF from every value before writing it, and skip the field if stripping empties it. IP/URL/key values are unaffected; only embedded newlines are removed.

## Tests

Added regression tests to src/lib/credentials.test.ts covering values with embedded \n, \r, and \r\n -- asserting the serialized output stays single-line-per-field and round-trips without injected keys. Fail on main before the fix, pass after.

Local run: the only failing credentials tests are the pre-existing Windows-only ones (file mode 0600 -> 0666 and the ~ path check); they pass on Linux CI. typecheck + prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Credential serialization now removes newline characters from saved values, helping prevent malformed output and injection issues.
  * Empty values created by sanitization are skipped instead of being written.

* **Tests**
  * Added coverage for credential serialization cases involving embedded line breaks to verify sanitized output remains valid.

### DIFF
--- a/src/lib/credentials.test.ts
+++ b/src/lib/credentials.test.ts
@@ -86,6 +86,38 @@ describe('serializeCredentials', () => {
     expect(text).toContain('api_key = sk');
     expect(text).not.toContain('api_url');
   });
+
+  it('strips newline characters from values to prevent INI injection', () => {
+    // A malicious apiUrl with embedded newlines could inject new key-value
+    // pairs or section headers into the credentials file. The serializer
+    // must strip \n and \r so the written file has exactly one value per
+    // field and no injected content parsed as separate keys/sections.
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):

PR #136, opened 2026-07-02
fix(steps): surface run-scoped step errors

### LINKED ISSUE #126: [Hackathon] "test steps --run-id" silently discards the per-step error text the API already returned

interferon0

RunStepDto carries error: string | null (the per-step failure text) and type ('action' | 'assertion') on GET /runs/{id}?includeSteps=true (src/lib/runs.types.ts:175-179), but the mapping into the CLI's step shape drops both (mapRunStepToCliTestStep, src/commands/test.ts:3731-3751, into CliTestStep which has neither field). So an agent that just ran a test and asks for the failing run's steps sees "status": "failed" with no reason — in text AND in --output json — and must download the full artifact bundle (video, screenshots, DOM) to read one string the CLI already had in memory. Carry error through as an additive JSON field and print an indented error: sub-line under failed rows in text mode; optionally carry type as stepType (the current renderer guesses assertion-ness from the action string, test.ts:8522-8523).

Additive change only (new optional fields, nothing renamed), mirroring the existing sub-line pattern used for targetUrl in the history table (test.ts:4164-4173). The data is already fetched in the same HTTP response — zero new wire calls, no new dependency. Tests: error surfaces in JSON and text for a failed step, absent on passing steps. PR to follow once assigned.

### Confirmations

JerryNee:
## Summary
- carry `RunStepDto.type` into `CliTestStep.stepType`
- carry per-step `RunStepDto.error` into JSON output for `test steps --run-id`
- render an indented `error:` sub-line below steps that include backend failure text

Fixes #126.

## Tests
- npm test -- src/commands/test.test.ts -t=run-id
- npm run lint
- npm run typecheck
- npm test
- npm run build

## Summary by CodeRabbit

* **New Features**
  * Test step output now includes step type details and surfaces step-specific error messages in both JSON and text formats.
  * Text output now displays an additional indented `error:` line under the failed step row when an error is present.

* **Bug Fixes**
  * Improved `--run-id` output validation to consistently report per-step fields, including correct failure-related information.
  * JSON output now preserves the exact per-step error text for failed steps.

* **Tests**
  * Added coverage for `--run-id` when a specific failed step includes an error string, plus updated dry-run sample assertions for failed assertion steps.

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -2522,11 +2522,81 @@ describe('runSteps', () => {
     // timestamps come from RunStepDto.createdAt
     expect(first.capturedAt).toBe('2026-06-01T10:00:05.000Z');
     expect(first.updatedAt).toBe('2026-06-01T10:00:05.000Z');
+    expect(first.stepType).toBe('action');
+    expect(first.error).toBeNull();
     // outcomeContributesToFailure: null when run passed (failedStepIndex is null)
     expect(first.outcomeContributesToFailure).toBeNull();
     // JSON output should be parseable with 2 items
-    const printed = JSON.parse(out[0]!) as { items: unknown[] };
+    const printed = JSON.parse(out[0]!) as {
+      items: Array<{ stepType?: string; error?: string | null }>;
+    };
     expect(printed.items).toHaveLength(2);
+    expect(printed.items[0]!.stepType).toBe('action');
+    expect(printed.items[0]!.error).toBeNull();
+  });
+
+  it('--run-id preserves per-step error text in JSON output', async () => {
+    const { credentialsPath } = makeCreds();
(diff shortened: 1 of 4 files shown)


zeshi-du:
Closing in favor of #167, which landed today with the same fields (`stepType` + `error` carried through the run-scoped mapping) plus two robustness pieces the text renderer needed: wire error text is almost always a multi-line Playwright stack, so the sub-line collapses newlines and caps at 200 chars to keep the table intact.

You were first to this gap, and one slice of your PR is still uncovered: the **dry-run sample enrichment** (`samples.ts` failed-step entry + its tests) — #167 didn't touch samples, so a follow-up PR with just that slice would be welcome and merges fast. 🙏

PR #137, opened 2026-07-02
fix(http): clear per-attempt timeout timers

### LINKED ISSUE #82: [Hackathon] Per-Attempt AbortSignal.timeout Timers Are Never Cleared (Timer/Memory Growth in Large Batches)

uisupersaiyan3

### Summary
 
Each request attempt mints a fresh `AbortSignal.timeout(this.requestTimeoutMs)`
and never cancels it once the request succeeds. The underlying timer and
signal live for their full duration (120 seconds), even when the request
itself completed in milliseconds.
 
---
 
### Root Cause
 
`AbortSignal.timeout(...)` has no built-in way to cancel the timer it
creates once the associated operation completes. The timer keeps running
until it either fires or the process exits, regardless of whether the
signal is still needed. Because a new signal is created per attempt rather
than reused or explicitly cleaned up, successful requests leave their timer
running for no benefit.

---
 
### Affected Code
 
| File | Lines | Behavior |
|---|---|---|
| `src/lib/http.ts` | 431 | Mints `AbortSignal.timeout(this.requestTimeoutMs)` per attempt |
| `src/lib/http.ts` | 433 | Composes the timeout signal via `AbortSignal.any`, with no corresponding cleanup on settle |
 
---
 
### Impact
 
For a single request, this is negligible: one uncollected timer for up to
120 seconds is not meaningful overhead.
 
For high-throughput usage, it accumulates. A `test ... --wait` batch can
dispatch up to 50 concurrent triggers, each followed by repeated poll GETs,
and each individual GET mints its own fresh 120-second timer regardless of
how quickly it resolves. Over the course of a long-running batch, this
results in thousands of live timers and signal objects accumulating in
memory at any given moment.
 
These timers are unref'd, so they do not block process exit. The impact is
memory growth during long-running batches, not a hang or a crash.

### Proposed Resolution
 
Replace `AbortSignal.timeout(...)` with an explicit `AbortController` paired
with `setTimeout`, and clear that timeout in a `finally` block once the
fetch settles:
 
- Create an `AbortController` and a `setTimeout` that calls `controller.abort()`
  at `requestTimeoutMs`.
- Pass `controller.signal` into the fetch call as before, composed with any
  caller-provided abort signal.
- In a `finally` block around the fetch, call `clearTimeout` unconditionally,
  so the timer is released immediately once the request settles, whether it
  succeeded, failed, or was retried.

### Confirmations

JerryNee:
## Summary
- replace per-attempt `AbortSignal.timeout(...)` with an explicit `AbortController` + clearable timer
- clear the attempt timeout after response bodies settle and before retry backoff sleeps
- add regression coverage proving settled attempt signals do not abort later

Fixes #82.

## Tests
- npm test -- src/lib/http.test.ts
- npm run typecheck
- npm run lint
- npm test
- npm run build

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed per-request timeout behavior so timeouts are cleared after successful responses and don’t interfere across retry/backoff periods.
  * Improved combined caller-abort and timeout handling to ensure the correct error type is surfaced when both could trigger.
  * Refined retry behavior for non-OK responses by honoring a clamped `Retry-After` value during error classification and retry decisions.
* **Tests**
  * Added coverage for per-attempt timeout clearing, retry signal handling, and timeout-vs-caller-abort error precedence.

### DIFF
--- a/src/lib/http.test.ts
+++ b/src/lib/http.test.ts
@@ -449,6 +449,52 @@ describe('HttpClient per-request timeout', () => {
     expect(callCount).toBe(1);
   });
 
+  it('clears the per-attempt timeout after a successful response body is read', async () => {
+    let capturedSignal: AbortSignal | undefined;
+    const fetchImpl = vi.fn(async (_input: unknown, init?: RequestInit) => {
+      capturedSignal = init?.signal ?? undefined;
+      return jsonResponse({ ok: true });
+    });
+    const client = new HttpClient({
+      baseUrl: 'https://api.example.com/api/cli/v1',
+      apiKey: 'sk-test',
+      fetchImpl: fetchImpl as unknown as typeof fetch,
+      sleep: () => Promise.resolve(),
+      random: () => 0,
+      requestTimeoutMs: 25,
+    });
+
+    await expect(client.get('/me')).resolves.toEqual({ ok: true });
+    expect(capturedSignal?.aborted).toBe(false);
+    await new Promise(resolve => setTimeout(resolve, 50));
+    expect(capturedSignal?.aborted).toBe(false);
+  });
+
+  it('clears a failed attempt timeout before retry backoff sleeps', async () => {
+    const attemptSignals: AbortSignal[] = [];
+    let calls = 0;
+    const fetchImpl = vi.fn(async (_input: unknown, init?: RequestInit) => {
+      if (init?.signal) attemptSignals.push(init.signal);
+      calls += 1;
+      if (calls === 1) return errorEnvelopeResponse(500, 'INTERNAL');
+      return jsonResponse({ ok: true });
+    });
+    const client = new HttpClient({
+      baseUrl: 'https://api.example.com/api/cli/v1',
+      apiKey: 'sk-test',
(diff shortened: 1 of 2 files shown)


zeshi-du:
This is approved in substance — replacing `AbortSignal.timeout()` with a cleared, unref'd timer is right (no timer pile-up across retries, no exit-latency tail), and the `effectiveSignal` tie-break fix for "caller aborts after the timeout wins" is a genuinely subtle catch. The tests for both are exactly what we'd want.

It just went CONFLICTING because #166 landed in the same `response.json()` catch block (non-JSON 200 → typed `INTERNAL` error). The resolution when you rebase:

```ts
} catch (err) {
  this.rethrowIfAbort(err, timeoutSignal, options.signal, requestId, effectiveSignal); // yours
  throw malformedResponseError(response, requestId, err); // #166's, instead of `throw err`
}
```

i.e. keep your five-arg `rethrowIfAbort`, but the fall-through now throws #166's `malformedResponseError` rather than the raw parse error. Rebase and it merges.

zeshi-du (approved):
Reviewed: scoped, tests included, CI green across the matrix.

Issue #94, opened 2026-07-01
[Hackathon] A 200 response with a non-JSON body crashes with an opaque SyntaxError and breaks the --output json error envelope


Davidson3556:
### Discord Username / User ID

Davidson

### What does this improvement do?

Make the HTTP layer handle a 200 OK response whose body is not JSON gracefully, with a typed error envelope, instead of crashing with an opaque SyntaxError (exit 1) that also breaks the --output json error contract.

Today, if the configured endpoint returns a 200 with a non-JSON body — a misconfigured TESTSPRITE_API_URL, a captive-portal / SSO login page, or a caching proxy that serves an HTML page with a 200 status — the CLI crashes with the raw response.json() parse error and no useful context.

### Details / implementation notes

$ node -e 'require("http").createServer((_,r)=>{r.writeHead(200,{"content-type":"text/html"});r.end("<!DOCTYPE html>...")}).listen(8791)' &
$ TESTSPRITE_API_KEY=sk-user-anything TESTSPRITE_API_URL=http://localhost:8791 \
    testsprite project list --output json
{
  "error": "Unexpected token '<', \"<!DOCTYPE \"... is not valid JSON"
}
# exit 1

### Confirmations

- [x] I have searched existing issues and this is not a duplicate.
- [x] I have provided my Discord identity above for reward coordination.

Davidson3556:
can you assign this to me

zeshi-du:
To claim an issue, comment `/assign` (the literal slash command) — the triage bot only reacts to that exact form, not free-text requests.

Davidson3556:
As the assigned owner of this issue, I've opened PR #166 with a fix: it maps a non-JSON `200` body to a typed `ApiError` envelope (carrying `requestId` + `details`) instead of leaking a raw `SyntaxError`, keeping exit 1 and restoring the `--output json` error contract. All checks pass locally (typecheck, lint, format, 1618 unit tests, e2e).

I notice #142 also addresses this issue — I'm happy to coordinate with the maintainers on whichever approach you prefer.

PR #142, opened 2026-07-03
fix: handle non-json 200 responses

naufalfx805-source:
Refs #94

Discord: npall_805

Summary:
- map HTTP 200 non-JSON success bodies to a typed UNAVAILABLE ApiError instead of leaking SyntaxError
- preserve abort/timeout classification during body reads
- add regression coverage for text/html 200 responses without retrying or leaking the HTML body

Validation:
- npm test -- src/lib/http.test.ts
- npm run typecheck
- npm run lint
- npm run build
- npx prettier --check src/lib/http.ts src/lib/http.test.ts
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when a request succeeds but returns non-JSON content.
  * `client.get()` now reports a clearer API error instead of exposing a raw parsing failure.
  * Added richer error details, including HTTP status and response content type, plus guidance for checking endpoint configuration.
  * Debug logging now follows the expected request → response → error sequence for this case.

### DIFF
--- a/src/lib/http.test.ts
+++ b/src/lib/http.test.ts
@@ -108,6 +108,29 @@ describe('HttpClient happy path', () => {
     expect(fetchImpl).toHaveBeenCalledTimes(1);
   });
 
+  it('maps a 200 non-JSON body to a typed UNAVAILABLE error', async () => {
+    const events: DebugEvent[] = [];
+    const fetchImpl = vi.fn(
+      async () =>
+        new Response('<!DOCTYPE html>', { status: 200, headers: { 'content-type': 'text/html' } }),
+    );
+    const client = makeClient(fetchImpl as unknown as typeof fetch, {
+      onDebug: e => events.push(e),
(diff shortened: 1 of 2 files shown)


Davidson3556:
please close this @zeshi-du , I have the issue assigned to me and I opened the issue first

zeshi-du:
Closing in favor of #166, which landed today covering the same gap (200-with-non-JSON-body → typed error instead of a raw SyntaxError). The taxonomy call went to `INTERNAL`/exit 1 — a captive portal or misconfigured endpoint won't fix itself, so an exit code scripts might blind-retry on (`UNAVAILABLE`/10) felt wrong. Two things from your version were genuinely better and are worth follow-ups if you want them: the **debug `error` event emission** on this path (#166 doesn't emit one), and the `details.contentType` normalization. Thanks — you were first to file this one.

Issue #86, opened 2026-07-01
[Hackathon] Redact presigned video/analysis URLs in text output


Andy00L:
### Discord Username / User ID

interferon0

### What does this improvement do?

test failure get and test result print videoUrl (and failureAnalysisUrl) as full presigned URLs in text mode. Those URLs carry their signature in the query string (bearer-capable, ~15 min TTL), so they leak into terminal scrollback and CI logs (e.g. testsprite test wait ... | tee run.log). This redacts the signature from the human text output while keeping the raw signed URL in --output json (the automation contract). The CLI already redacts this class of URL elsewhere via redactUrl, so it closes an inconsistency.

### Details / implementation notes

One-line change per site in src/commands/test.ts (renderFailureContextText, renderResultText), wrapping the URL with the existing redactUrl helper (strips query + userinfo, keeps scheme/host/path). Two regression tests assert the signature query is stripped from text while host/path remain; --output json unchanged. Full gate green (tsc/eslint/prettier, 1710 tests, coverage 88.9%). PR to follow once assigned.

### Confirmations

- [x] I have searched existing issues and this is not a duplicate.
- [x] I have provided my Discord identity above for reward coordination.

zeshi-du:
Triage: **declined** — in text mode those URLs are exactly what a human needs to click, and stripping the signature turns them into dead links; the short TTL is the leak mitigation by design. (Also could not find the `redactUrl` precedent this cites in current source.) A `--debug`-stream-only redaction would be considered as a separate proposal.

zeshi-du:
Closing as not-planned per the 2026-07-03 triage decision (also applied to PR #143): text-mode URLs exist for a human to click, and a presigned S3 URL with its signature stripped is a dead link (403). Redaction did land where it belongs — error/diagnostic paths (#60) — but success-path text output stays clickable by design. `--output json` remains verbatim for machine consumers.

PR #143, opened 2026-07-03
fix: redact presigned urls in text output

naufalfx805-source:
Refs #86

Discord: npall_805

Summary:
- redact query strings, fragments, and userinfo from video/failure-analysis URLs in text renderers
- keep --output json unchanged so automation still receives raw presigned URLs
- add regression coverage for test result and failure get text/json paths

Validation:
- npm test -- src/commands/test.test.ts
- npm run typecheck
- npm run lint
- npm run build
- npx prettier --check src/commands/test.ts src/commands/test.test.ts
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI output so sensitive parts of video and failure-analysis links are no longer shown in text mode.
  * Preserved full link details in JSON output while making human-readable summaries cleaner and safer.
  * Updated test coverage to match the new link formatting behavior.

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -2737,6 +2737,8 @@ describe('runResult', () => {
     expect(seen[0]).toContain('/tests/test_fe/result');
     expect(got).toEqual(RESULT_FAILED);
     expect(JSON.parse(out[0]!)).toEqual(RESULT_FAILED);
+    expect(out[0]).toContain('video/run_abc.mp4?X');
+    expect(out[0]).toContain('analysis/run_abc.json?X');
   });
 
   it('text mode for a failed run highlights failureKind + failedStepIndex up top', async () => {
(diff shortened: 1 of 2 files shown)


zeshi-du:
Thanks for the careful implementation — the tests are thorough and the JSON-mode passthrough is exactly right. But we're going to pass on this one as a matter of product policy rather than code quality: it was decided on issue #86 (not-planned) that **text mode keeps full presigned URLs** — the text-mode URL is there for a human to click, and a presigned S3 URL with its signature stripped is a dead link (S3 returns 403 without the query string). Redaction in *error/diagnostic* paths did land (#60 territory) because there the URL is context, not a clickable deliverable — but success-path text output stays clickable by design.

Closing per that decision — genuinely appreciate the contribution, and hope one of the open `accepted` issues catches your interest.

PR #196, opened 2026-07-06
fix(bundle): atomic re-commit via per-entry aside (compatible with #162)

SahilRakhaiya05:
## Summary

Re-running `test failure get --out <dir>` can delete `meta.json` mid-commit; a failed re-commit then leaves the directory without a usable bundle.

This redesigns `commitBundle` to be **compatible with #162** (`isBundleOwnedEntry`): `--out` may contain the user's unrelated files, so a whole-directory swap is not safe.

## Approach

1. Stage the complete bundle in `<dir>/.tmp/` (unchanged)
2. Move only **bundle-owned** entries aside to a sibling `.<dir>.aside.<uuid>/` directory
3. Install new artifacts from `.tmp/`; `meta.json` lands last (completion signal)
4. On failure, restore aside entries (rollback) - prior bundle survives
5. On success, best-effort aside cleanup (cleanup failure does not roll back)

Foreign files (e.g. `notes.txt`, `src/`) are never touched.

## Testing

- `npx vitest run src/lib/bundle.commit.test.ts` - rollback, foreign-file preservation, aside cleanup
- `npx vitest run src/lib/bundle.test.ts -t "commit sweep"` - existing #162 ownership guard

Split from #8 per review feedback.

## Summary by CodeRabbit

* **New Features**
  * Improved bundle installation for `--out` directories that already include unrelated files.

* **Bug Fixes**
  * Preserves unrelated files during bundle updates.
  * Adds safer rollback behavior when a commit/install step fails, keeping the prior bundle intact.
  * Improves cleanup of temporary install state after both success and failure scenarios.

* **Tests**
  * Added coverage for rollback and cleanup edge cases during bundle commit operations.


zeshi-du (changes requested):
This is very close — the aside/rollback design is a real improvement over the old commit path (which had no rollback at all), and the #162 foreign-file contract is correctly preserved. One ordering issue before it can land:

**`meta.json` must be the FIRST entry moved aside, not wherever `readdir` happens to place it.**

The old commit path removed the old `meta.json` first, deliberately: its presence is the bundle's completion signal (§3/§7.3 — "no meta → bundle absent or mid-write, refuse to consume"). In this version the aside loop walks `readdir(dir)` order, which is platform-arbitrary. If `steps/` is moved aside before `meta.json`, there's a window where a concurrent reader sees a present `meta.json` (⇒ "bundle complete") pointing at steps that are already gone — exactly the torn read the meta-first ordering existed to prevent.

Fix is one line: `await asideIfPresent('meta.json')` before the loop (and skip it inside the loop). Install-side ordering is already right (meta lands last).

Nit, non-blocking: `asideDir` lives in `dirname(dir)`, so a read-only parent with a writable `--out` dir now fails the commit where it previously succeeded. The failure is caught and rolled back cleanly, so I'm fine with it — just noting the behavior change.

Happy to merge as soon as the meta-first ordering is in.

jangjos-128:
Hi @SahilRakhaiya05 — so sorry for the disruption here. 🙏 While we were publishing a new CLI release, a hiccup in our process unintentionally closed this PR. To be clear: **your contribution wasn't rejected**, and nothing on your end caused this.

Issue #79, opened 2026-07-01
[Hackathon] invalid --since and missing --password-file crash instead of returning VALIDATION_ERROR


crypticsaiyan:
### Discord Username / User ID

uisupersaiyan3

### Summary
 
Two categories of invalid user input currently produce unhandled exceptions
(raw Node stack traces, exit code `1`) instead of the CLI's standard typed
validation error (exit code `5`):
 
1. An extremely large `--since` duration overflows the valid JavaScript date
   range and crashes during date conversion.
2. A non-existent `--password-file` path during project creation or update
   crashes with an unguarded filesystem error.
In both cases, input that a user could plausibly supply by mistake bypasses
the CLI's validation-error handling entirely and surfaces as an
unhandled exception instead.
 
---
 
### Affected Code
 
| File | Function | Issue |
|---|---|---|
| `src/commands/test.ts` | `parseDuration` | No bounds check on the parsed date before calling `.toISOString()` |
| `src/commands/project.ts` | Password file handling (create/update) | `readFileSync` call is unguarded; missing file throws raw `ENOENT` |
 
 

### Details / implementation notes

issue fixed in PR #27 

### Confirmations

- [x] I have searched existing issues and this is not a duplicate.
- [x] I have provided my Discord identity above for reward coordination.

zeshi-du:
Status correction after re-auditing the code: PR #27 fixed only the --since half of this issue (parseDuration overflow). The --password-file half is **still live on main** — project.ts calls readFileSync(opts.passwordFile) bare at three sites, so a missing file still crashes with a raw ENOENT instead of a VALIDATION_ERROR envelope. Keeping this open scoped to the password-file half; the split-out password-file guard from #248's rework (see the review there) is the natural vehicle. Sorry for the earlier over-claim.

naufalfx805-source:
/assign

Picking up the remaining `--password-file` half, scoped per @zeshi-du's 2026-07-18 status correction.

Planning to cover the `project create` and `project update` call sites (`project.ts:213`, `project.ts:329`) and leave `project auto-auth` (`project.ts:579`) to #282, which is already in progress and covers that site alongside its three sibling flags. The guard will be a shared helper taking the flag name, so #282 can adopt it for `--credential-file` / `--client-secret-file` / `--refresh-token-file` without rework.

PR #204, opened 2026-07-06
fix(project): validate missing password files

Lexiie:
## Summary

Addresses the `--password-file` part of #79.

`testsprite project create/update --password-file <path>` previously read the file directly with `readFileSync`. If the file was missing or unreadable, the raw filesystem error escaped instead of the CLI's typed `VALIDATION_ERROR` envelope.

This PR keeps dry-run behavior unchanged, but on real create/update paths converts password-file read failures into exit 5 validation errors before any network work.

## Changes

- Add `readPasswordFile()` for project create/update password-file reads.
- Map missing files to `VALIDATION_ERROR` with `field: "password-file"`.
- Map permission and other read failures to typed validation errors too.
- Add regression coverage for create and update ensuring missing password files fail before fetch.

## Verification

- `npm run format:check`
- `npm run typecheck`
- `npx vitest run src/commands/project.test.ts`
- `npx eslint src/commands/project.ts src/commands/project.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for `--password-file` in project create/update flows.
  * Missing, unreadable, or inaccessible password files now fail fast with a clear local validation error instead of continuing.
  * Added coverage to ensure invalid password-file paths are rejected before any network request is made.

### DIFF
--- a/src/commands/project.test.ts
+++ b/src/commands/project.test.ts
@@ -664,6 +664,39 @@ describe('runCreate', () => {
     ).rejects.toMatchObject({ exitCode: 5, code: 'VALIDATION_ERROR' });
     expect(fetchImpl).not.toHaveBeenCalled();
   });
+
+  it('rejects a missing --password-file with VALIDATION_ERROR (exit 5), no network', async () => {
+    const { credentialsPath } = makeCreds();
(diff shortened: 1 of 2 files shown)


Lexiie:
Closing this to avoid duplicating the existing password-file guard work in #61. I missed that #61 was still the maintainer-preferred path for the remaining password-file half after #27 merged only the --since overflow guard.

PR #205, opened 2026-07-06
fix(bundle): preserve root output path

Lexiie:
## Summary

`resolveBundleDir('/')` currently strips the trailing slash before checking whether the path is absolute. For the filesystem root path, that turns `/` into an empty string, so the helper resolves it against `process.cwd()` instead of preserving `/`.

This PR keeps the root path intact while preserving the existing behavior for ordinary trailing slashes like `/tmp/bundle/`.

## Changes

- Only strip a trailing slash when the raw path has more than one character.
- Add regression coverage that `resolveBundleDir('/')` returns `/`.

## Verification

- `npm run format:check`
- `npm run typecheck`
- `npx vitest run src/lib/bundle.test.ts`
- `npx eslint src/lib/bundle.ts src/lib/bundle.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the filesystem root path (`/`) when resolving bundle directories, avoiding it being reduced to an empty path.
  * Continued trimming trailing slashes for non-root paths as expected.

* **Tests**
  * Added coverage to verify root paths remain unchanged during bundle directory resolution.


jangjos-128:
Hi @Lexiie — so sorry for the disruption here. 🙏 While we were publishing a new CLI release, a hiccup in our process unintentionally closed this PR. To be clear: **your contribution wasn't rejected**, and nothing on your end caused this.

Issue #186, opened 2026-07-06
[Hackathon] Add structured debug logging to silent catch blocks


farhank15:
### Discord Username / User ID

getmawa_22

### What does this improvement do?

Multiple catch blocks across the CLI silently swallow errors with no
logging or trace output, making debugging difficult. When a feature
silently degrades (e.g., duplicate-name advisory, skill nudge,
bundle stream verification), the operator has no way to know an
error occurred — the CLI just returns a confusing result.

This adds structured debug logging (gated behind `--debug` / `--verbose`)
to these catch blocks so developers and support can trace suppressed
errors without changing behavior or exit codes.

### Details / implementation notes

Found ~12 catch blocks across:
- src/commands/test.ts (lines 634, 4429, 4440, 4727, 5739)
- src/lib/skill-nudge.ts (lines 63, 130)
- src/lib/bundle.ts (line 606)
- src/commands/init.ts (lines 43, 290, 316)
- src/lib/dry-run/fetch.ts (line 41)
- src/lib/facade.ts (lines 78, 95)

For each location, inject a conditional stderr log when `opts.debug`
or `opts.verbose` is set, e.g.:
  ```
  if (opts.debug) {
    stderr(`[debug] <context>: ${err instanceof Error ? err.message : String(err)}`);
  }
  ```

Changes are additive — no existing behavior is altered, no exit codes
change, and the default (no --debug) output is byte-identical.
Estimated < 100 lines changed across ~12 files.

Tests: verify that with --debug the catch path prints a stderr line
and without --debug the output is unchanged (byte-identical snapshots).

### Confirmations

- [x] I have searched existing issues and this is not a duplicate.
- [x] I have provided my Discord identity above for reward coordination.

zeshi-du:
Scope bookkeeping so this doesn't get closed prematurely: the issue enumerates ~12 silent catch sites across 6 files. So far exactly one is covered (init.ts whoami fallback, merged via #223); a second PR (#229) targeted update-check.ts, which isn't on this issue's list, and got a rework request. Everything else here — test.ts ×5, skill-nudge.ts ×2, bundle.ts, the remaining init.ts sites, dry-run/fetch.ts, facade.ts ×2 — is still open for contributors. The pattern to follow is #223: gate strictly on the parsed --debug option, stderr only, with a byte-identical-without-debug test.

PR #209, opened 2026-07-07
fix(init): debug log whoami fallback

Lexiie:
## Summary

Covers one `src/commands/init.ts` catch path from #186.

After `setup` successfully configures credentials, it runs a display-only `whoami` lookup to fill the final summary. If that lookup fails, setup intentionally continues with a placeholder summary. That behavior is fine, but the failure was completely silent, so `--debug` users could not tell why the summary fell back to no email / production / empty scopes.

This PR keeps default output and exit behavior unchanged, but emits a debug-only stderr line when the display-only identity lookup fails.

## Changes

- Capture the swallowed `runWhoami` error in `runInit`.
- Emit `[debug] setup identity lookup failed after configure: ...` only when `--debug` is set.
- Add a regression test that JSON stdout remains parseable while the debug line goes to stderr.

## Verification

- `npm run format:check`
- `npm run typecheck`
- `npx vitest run src/commands/init.test.ts`
- `npx eslint src/commands/init.ts src/commands/init.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initialization behavior when identity lookup fails in debug mode.
  * JSON output now stays valid even if the setup identity check encounters an error.
  * Added a debug message to help explain identity lookup failures without interrupting the setup summary.


jangjos-128:
Hi @Lexiie — so sorry for the disruption here. 🙏 While we were publishing a new CLI release, a hiccup in our process unintentionally closed this PR. To be clear: **your contribution wasn't rejected**, and nothing on your end caused this.

PR #211, opened 2026-07-07
test(test): guard status filter validation order

Lexiie:
## Summary
- Add a focused `test list --status junk` regression test that runs without credentials.
- Pin the expected local `VALIDATION_ERROR` behavior so invalid status filters cannot regress into `AUTH_REQUIRED` if validation order changes.

## Changes
- Covers the `runList` command-parser path for `--status` validation before credentials/client setup.

## Verification
- `rtk vitest run src/commands/test.test.ts -t "rejects --status=junk before reading credentials"`
- `rtk npx eslint src/commands/test.test.ts --fix-dry-run --format stylish`
- `rtk npx eslint src/commands/test.test.ts`
- `rtk npm run typecheck`

Fixes #210

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for `test list` so invalid status values are rejected immediately with a clear validation error, before any credentials are read or network requests are made.
* **Tests**
  * Added coverage for invalid status handling in the CLI.

PR #222, opened 2026-07-09
test(test): guard status filter validation order

### LINKED ISSUE #210: [Hackathon] guard test list status validation before auth

lexiisme

`testsprite test list --status <invalid>` should always fail as local input validation before credentials, network, or filesystem work.

Current `upstream/main` already validates the status filter before constructing the client, but there was no focused regression test pinning that order. Without coverage, a future refactor could accidentally move status validation after client creation again, causing machines without configured credentials to see `AUTH_REQUIRED` instead of the expected `VALIDATION_ERROR` for bad user input.

Expected behavior:

- `test list --project project_alice --status junk` should reject locally with `VALIDATION_ERROR`.
- The command should not read credentials or construct the HTTP client before rejecting the invalid status token.
- This keeps status-filter validation aligned with the existing local validation behavior for missing `--project`, invalid `--type`, invalid `--created-from`, and invalid pagination flags.

Proposed fix:

- Add a focused `runList` command-parser regression test with no injected credentials.
- Assert invalid `--status` rejects with `VALIDATION_ERROR` so validation order remains protected.
- No runtime code change is needed on current `main`; this is a coverage guard for an already-correct implementation order.

### Confirmations

Lexiie:
## Summary
- Resubmission of #211 after the release-process hiccup closed it without merge.
- Add a focused `test list --status junk` regression test that runs without credentials.
- Pin the expected local `VALIDATION_ERROR` behavior so invalid status filters cannot regress into `AUTH_REQUIRED` if validation order changes.

## Changes
- Covers the `runList` command-parser path for `--status` validation before credentials/client setup.

## Verification
- `rtk vitest run src/commands/test.test.ts -t "rejects --status=junk before reading credentials"`
- `rtk npx eslint src/commands/test.test.ts --fix-dry-run --format stylish`
- `rtk npx eslint src/commands/test.test.ts`
- `rtk npm run typecheck`

Fixes #210

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation so invalid `--status` values are rejected immediately when listing tests, with a clear validation error.
  * Prevents unnecessary follow-up actions after an invalid command is entered.

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -464,6 +464,14 @@ describe('runList', () => {
     ).rejects.toMatchObject({ code: 'VALIDATION_ERROR' });
   });
 
+  it('rejects --status=junk before reading credentials', async () => {
+    const test = createTestCommand();
+    disableExits(test);
+    await expect(
+      test.parseAsync(['list', '--project', 'project_alice', '--status', 'junk'], { from: 'user' }),
+    ).rejects.toMatchObject({ code: 'VALIDATION_ERROR' });
+  });
+
   it('rejects --page-size=0 locally with VALIDATION_ERROR (no network call)', async () => {
     const { credentialsPath } = makeCreds();
     const fetchImpl = makeFetch(() => {


zeshi-du (approved):
Merging as a harmless additive test: note the ordering fix itself (validate --status before client/auth setup) already shipped via the #152 backport, whose test asserts the stronger property (fetch never called). This one adds the CLI-wiring angle via createTestCommand + parseAsync, which is a distinct layer. Thanks @Lexiie!

PR #223, opened 2026-07-09
fix(init): debug log whoami fallback

Lexiie:
## Summary

Resubmission of #209 after the release-process hiccup closed it without merge.

Covers one `src/commands/init.ts` catch path from #186.

After `setup` successfully configures credentials, it runs a display-only `whoami` lookup to fill the final summary. If that lookup fails, setup intentionally continues with a placeholder summary. That behavior is fine, but the failure was completely silent, so `--debug` users could not tell why the summary fell back to no email / production / empty scopes.

This PR keeps default output and exit behavior unchanged, but emits a debug-only stderr line when the display-only identity lookup fails.

## Changes

- Capture the swallowed `runWhoami` error in `runInit`.
- Emit `[debug] setup identity lookup failed after configure: ...` only when `--debug` is set.
- Add a regression test that JSON stdout remains parseable while the debug line goes to stderr.

## Verification

- `rtk vitest run src/commands/init.test.ts -t "debug log whoami fallback|whoami fallback|debug"`
- `rtk npx eslint src/commands/init.ts src/commands/init.test.ts --fix-dry-run --format stylish`
- `rtk npx eslint src/commands/init.ts src/commands/init.test.ts`
- `rtk npm run typecheck`

Refs #186

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved init handling when identity lookup fails during setup, so the process still completes and prints a valid summary.
  * Added debug logging for lookup failures, while keeping JSON output clean and parseable.
  * Added coverage to verify the initialized status is still returned and debug messages appear only on stderr.

### DIFF
--- a/src/commands/init.test.ts
+++ b/src/commands/init.test.ts
@@ -231,6 +231,45 @@ describe('runInit — happy path (interactive)', () => {
     expect(agent.skills).toContain('testsprite-verify');
     expect(agent.skills).toContain('testsprite-onboard');
   });
+
+  it('debug mode reports a display-only whoami lookup failure without corrupting JSON stdout', async () => {
+    const { captured, deps } = makeCapture();
+    let callCount = 0;
+    const fetchMock = vi.fn(async () => {
+      callCount += 1;
+      if (callCount === 1) {
+        return new Response(JSON.stringify(ME), {
+          status: 200,
+          headers: { 'content-type': 'application/json' },
+        });
+      }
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Debug line is strictly gated behind --debug, lands on stderr only, and this is the first test coverage for that silent catch. Matches the scoped direction from #186. Thanks @Lexiie!

PR #233, opened 2026-07-10
fix(config): normalize empty/whitespace TESTSPRITE_PROFILE to unset

0xshalah:
One-line fix: wraps env.TESTSPRITE_PROFILE with normalizeEnvVar() so an
empty/whitespace-only value is treated as unset and resolves to the "default"
profile — consistent with TESTSPRITE_API_URL and TESTSPRITE_API_KEY.

3 new unit tests added (empty, whitespace-only, blank-with-credentials).
eslint + prettier + tsc all clean locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration loading when the profile environment setting is empty or contains only whitespace.
  * Automatically falls back to the default profile in these cases, including its configured credentials.

* **Tests**
  * Added coverage for empty, whitespace-only, and default profile configuration scenarios.

### DIFF
--- a/src/lib/config.test.ts
+++ b/src/lib/config.test.ts
@@ -55,6 +55,23 @@ describe('loadConfig', () => {
     ).toBe('option-profile');
   });
 
+  it('treats empty TESTSPRITE_PROFILE as unset (falls back to default)', () => {
+    expect(loadConfig({ env: { TESTSPRITE_PROFILE: '' }, credentialsPath }).profile).toBe(
+      'default',
+    );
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Surgical fix reusing the existing normalizeEnvVar helper already applied to the sibling env vars in the same function, with the three targeted tests. The red gate check is a repo-wide CI infra issue we just fixed on main — unrelated to this PR. Thanks @0xshalah!

PR #234, opened 2026-07-10
fix(init): skip API key prompt when credentials already exist

### LINKED ISSUE #206: [Hackathon] setup should not re-prompt for API key when credentials already exist

hempun10

Right now, running `testsprite setup` **always asks for your API key**, even if you already configured one and it is saved in `~/.testsprite/credentials`.

That is confusing because:

- Other commands (`auth status`, `project list`, `test run`, etc.) already use the saved key without asking again.
- Re-running `setup` usually means "install or refresh the agent skill," not "type my API key again."
- New users cannot tell whether their key was saved or if something went wrong.

This improvement would make `setup` **smarter on repeat runs**: detect an existing profile, show what is already configured, and let the user **keep it**, **update the key**, or **set up a different profile** — instead of blindly prompting every time.

**Note:** API keys are stored **globally** on your machine (`~/.testsprite/credentials`), not per project. This issue is about better UX for that global store, not about linking keys to individual TestSprite projects.

#### Steps to reproduce (today)

```bash
# First run — enter your API key when prompted
testsprite setup

# Confirm the key was saved
testsprite auth status   # works, no prompt

# Run setup again
testsprite setup         # asks for API key again
```

#### What happens today

- `setup` calls `runConfigure` in `src/commands/auth.ts`, which always prompts for the API key in interactive mode.
- It **does** read an existing profile, but only to inherit `api_url` — not to skip the key prompt.
- Tests in `src/commands/auth.test.ts` expect this (e.g. re-configure keeps the endpoint but still prompts for the key).

So this is **intentional today**, but it feels wrong for most users.

#### What we would like instead

When `setup` runs and the active profile already has a valid saved key:

1. **Show a short summary** — profile name, endpoint, masked key hint (e.g. last 4 chars), email from `GET /me` if available.
2. **Offer a simple choice** (interactive text mode):
   - **Keep existing credentials** (default) → skip key prompt, continue to skill install (unless `--no-agent`)
   - **Update API key** → prompt for a new key, validate, overwrite profile
   - **Cancel** → exit 0 with a clear message

For non-interactive / CI use, add something like:

- `--skip-if-configured` — if profile exists and key validates, skip the prompt and only run the skill install step
- Existing flags still work: `--api-key`, `--from-env`, `--no-agent`, `--yes`

#### Suggested acceptance criteria

- [ ] Re-running `testsprite setup` with a valid saved profile does **not** ask for the API key unless the user chooses to update it.
- [ ] `testsprite auth status` and `testsprite setup` (keep-existing path) resolve the **same** profile/key.
- [ ] `--output json` never blocks on an interactive prompt (same rule as today).
- [ ] `--dry-run` still makes no network calls and writes no credentials.
- [ ] Invalid or expired saved keys still prompt to enter a new one (with a clear message).
- [ ] Unit tests cover: first-time setup, repeat setup (keep), repeat setup (update), `--skip-if-configured`, non-interactive guardrails.

#### Code pointers

| File | Relevance |
|------|-----------|
| `src/commands/auth.ts` — `runConfigure()` | Where the key prompt lives |
| `src/commands/init.ts` — `runInit()` / `createSetupCommand()` | Orchestrates setup (configure + skill install) |
| `src/lib/credentials.ts` — `readProfile()` | Reads `~/.testsprite/credentials` |
| `src/lib/config.ts` — `loadConfig()` | How other commands resolve the active key |

#### Related issues (not duplicates)

- #30 / PR #31 — prompts moved to stderr (different problem)
- #73 / PR #183 — `doctor` checks setup health (workaround, not a fix)
- #147 — endpoint validation before key check
- #201 — skill text still references old `auth configure` flow

No PR yet — happy to implement once assigned.

### Confirmations

Tasfia-17:
## Summary

Running `testsprite setup` always re-prompts for the API key even when
credentials are already saved in `~/.testsprite/credentials`. This makes
re-running setup (a common pattern for refreshing the agent skill in
dotfiles and CI bootstraps) unnecessarily disruptive.

This PR adds `--skip-if-configured` to `testsprite setup` (and the
deprecated `init` alias). When the active profile already has a saved
API key and no explicit key source (`--api-key` or `--from-env`) is
given, the interactive prompt is skipped and the existing key is reused.

## Changes

- `src/commands/auth.ts` -- Added `skipIfConfigured` to `ConfigureOptions`. `runConfigure` short-circuits with `status: already_configured` before prompting when the flag is set and a saved key exists.
- `src/commands/init.ts` -- Added `--skip-if-configured` flag to `addSetupOptions`. `runInit` computes `skipWillApply` and relaxes the non-interactive guard and the `--output json` guard when skip applies. Cleared when `--api-key` is explicitly given so explicit keys always overwrite.
- `src/commands/auth.test.ts` -- 4 new test cases.
- `src/commands/init.test.ts` -- 4 new test cases.
- `test/__snapshots__/help.snapshot.test.ts.snap` -- Updated snapshot.

## Behaviour

- Re-running `testsprite setup` with a valid saved profile skips the key prompt.
- `--skip-if-configured` with `isTTY=false` works in CI without exit 5.
- `--api-key` always overwrites regardless of the flag.
- `--from-env` always overwrites regardless of the flag.
- `--dry-run` is unaffected (no network, no writes).
- `--output json` works without the prompt-corruption guard firing.

## Tests

All 1868 tests pass. Coverage stays above 80%.

Closes #206

## Summary by CodeRabbit

* **New Features**
  * Added a `--skip-if-configured` option to setup and initialization commands.
  * Existing saved credentials can now be reused without prompting or network validation.
  * Non-interactive and JSON modes can complete successfully when credentials are already configured.
  * Explicit API keys and environment-provided keys continue to override saved credentials.

* **Bug Fixes**
  * Prevented unnecessary failures in non-interactive setup when valid saved credentials are available.

### DIFF
--- a/src/commands/auth.test.ts
+++ b/src/commands/auth.test.ts
@@ -1224,3 +1224,90 @@ describe('createAuthCommand surface', () => {
     expect(err.exitCode).toBe(3);
   });
 });
+
+// ---------------------------------------------------------------------------
+// runConfigure -- skipIfConfigured
+// ---------------------------------------------------------------------------
+
+describe('runConfigure -- skipIfConfigured', () => {
+  it('skips the prompt and returns early when credentials already exist', async () => {
+    const { capture, deps } = makeCapture();
+    // Write a saved key first.
+    writeProfile('default', { apiKey: 'sk-existing' }, { path: credentialsPath });
+    const prompt = { secret: vi.fn(async () => 'sk-new') };
+    const fetchImpl = vi.fn();
+
+    await runConfigure(
+      { profile: 'default', output: 'text', debug: false, fromEnv: false, skipIfConfigured: true },
+      {
+        ...deps,
+        credentialsPath,
+        prompt,
+        fetchImpl: fetchImpl as unknown as AuthDeps['fetchImpl'],
+      },
+    );
+
+    // Prompt must never have fired.
+    expect(prompt.secret).not.toHaveBeenCalled();
+    // No network call -- we never validated or wrote a key.
+    expect(fetchImpl).not.toHaveBeenCalled();
+    // The saved key must be untouched.
+    expect(readProfile('default', { path: credentialsPath })?.apiKey).toBe('sk-existing');
+    // Output indicates already_configured.
+    expect(capture.stdout.join('\n')).toContain('already configured');
+  });
+
+  it('emits already_configured status in JSON mode', async () => {
+    const { capture, deps } = makeCapture();
+    writeProfile('default', { apiKey: 'sk-saved' }, { path: credentialsPath });
+    const fetchImpl = vi.fn();
+
+    await runConfigure(
+      { profile: 'default', output: 'json', debug: false, fromEnv: false, skipIfConfigured: true },
+      { ...deps, credentialsPath, fetchImpl: fetchImpl as unknown as AuthDeps['fetchImpl'] },
+    );
+
+    expect(fetchImpl).not.toHaveBeenCalled();
+    const parsed = JSON.parse(capture.stdout.join(''));
+    expect(parsed).toMatchObject({ profile: 'default', status: 'already_configured' });
+  });
+
+  it('proceeds normally when no credentials exist and skipIfConfigured is true', async () => {
+    const { deps } = makeCapture();
+    // No pre-existing profile -- skip has no effect, should fall through to prompt.
+    const prompt = { secret: vi.fn(async () => 'sk-new') };
+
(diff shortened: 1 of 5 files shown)


zeshi-du (approved):
Right shape for #206: opt-in via --skip-if-configured rather than a default behavior change, --api-key always wins, --from-env unaffected, profile-scoped lookup, and the non-interactive guards relax only when the flag is set and a key exists — with 8 tests covering each documented behavior. Thanks @Tasfia-17!

Issue #109, opened 2026-07-01
[Hackathon] Prevent concurrent credential profile writes from losing updates


SahilRakhaiya05:
### Discord Username / User ID

uxdev

### What does this improvement do?

Fix a credentials reliability issue where multiple TestSprite CLI processes can update the credentials file at the same time and accidentally overwrite each other’s changes.

Currently, `writeProfile` and `deleteProfile` perform read-modify-write operations on the credentials file. The final file rename is atomic, but the full read-modify-write sequence is not protected. If two CLI processes run concurrently, for example two terminals running `testsprite setup` for different profiles, each process can read the same old credentials snapshot. The last process to write wins, which can silently drop the other profile update.

This improvement adds cross-process locking around credential writes and deletes so only one process can modify the credentials file at a time. It preserves multiple concurrent profile updates and prevents credentials data loss.

### Details / implementation notes

This issue tracks the credentials write-locking fix implemented in PR #32.

Implementation approach:

* Wrap `writeProfile` and `deleteProfile` in a shared `withCredentialsLock()` helper.
* Add an exclusive lock file at `{credentialsPath}.lock`.
* Use exclusive lock acquisition so only one CLI process can enter the credentials read-modify-write section at a time.
* Store lock ownership metadata such as PID and timestamp.
* Add stale-lock recovery for abandoned locks when the owning process is dead or the lock has exceeded the configured age limit.
* Use bounded retry behavior with short backoff so contending processes wait instead of immediately failing.
* Ensure incomplete or partially written lock files are not incorrectly treated as stale.
* Clean up the lock file after successful write and delete operations.
* Preserve existing atomic credentials file write behavior.

Acceptance criteria:

* [ ] Concurrent `writeProfile` calls for different profiles preserve both updates.
* [ ] Concurrent `writeProfile` and `deleteProfile` operations do not corrupt the credentials file.
* [ ] Only one process can modify the credentials file at a time.
* [ ] Lock files are removed after successful `writeProfile` operations.
* [ ] Lock files are removed after successful `deleteProfile` operations.
* [ ] Stale locks from dead processes can be reclaimed safely.
* [ ] Fresh or incomplete lock files are not incorrectly reclaimed by another process.
* [ ] Lock acquisition has bounded retry behavior instead of hanging forever.
* [ ] First-time setup works when the credentials directory does not exist yet.
* [ ] Regression tests spawn concurrent writers and verify that multiple profile updates survive.
[...]

Issue #77, opened 2026-07-01
[Hackathon] Concurrent Profile Writes Silently Drop Updates (No File Lock)


crypticsaiyan:
### Discord Username / User ID

uisupersaiyan3

### Summary
 
`writeProfile` and `deleteProfile` implement a read-whole-file →
mutate-in-memory → `renameSync` cycle. The atomic rename guarantees the
credentials file is never left in a partially written state, but it does not
prevent a lost update: when two writers overlap, both read the same base
file, and whichever write completes last silently overwrites the other's
change.
 
There is currently no synchronization mechanism protecting this cycle. This
is a deterministic race condition, not an intermittent edge case.
 
---
 
### Affected Code
 
| File | Lines | Function(s) |
|---|---|---|
| `src/lib/credentials.ts` | 106–146 | `writeProfile`, `deleteProfile`, `writeCredentialsAtomic` |
| `src/commands/auth.ts` | 171 | Calls `writeProfile` |
| `src/commands/auth.ts` | 261 | Calls `deleteProfile` |
 
---
 
### Impact
 
- Running `testsprite setup --profile a` and `testsprite setup --profile b`
  concurrently, or executing two CI jobs that share `$HOME`, can result in
  one profile's freshly written API key being silently discarded.
- The failure mode is silent: no error is raised and no non-zero exit code
  is returned. The affected credential is simply absent on the next read,
  which makes the issue difficult to detect and diagnose after the fact.
- The `.tmp.${process.pid}` suffix (line 142) prevents temp-file name
  collisions between concurrent writers but does not address the underlying
  read-modify-write race on the destination file.
---
 
### Root Cause
 
The atomicity of the rename operation ensures the file is never corrupted
mid-write, but it does not determine which writer's in-memory snapshot
prevails. Neither writer is aware of the other's changes at the time it
writes.
 
---
 
### Supporting Evidence
 
- No advisory locking mechanism exists in the codebase. A search for
  `flock`, `O_EXCL`, and common lockfile patterns returns no matches under
  `src/`.
- No existing test exercises concurrent write behavior.

### Proposed Resolution
 
Introduce an advisory lock around the full read-modify-write cycle for every
credential mutation, rather than around the final rename alone:
 
- Acquire an `O_EXCL` lockfile at `{credentials-path}.lock` prior to reading
  the credentials file, and release it after the write completes.
- Implement bounded retry with backoff while waiting to acquire the lock.
[...]

PR #244, opened 2026-07-13
Recovered: fix(credentials): serialize profile writes with cross-process file locking (#32 by @SahilRakhaiya05)

jangjos-128:
## Summary

writeProfile and deleteProfile perform read-modify-write on `~/.testsprite/credentials`. The final `rename` is atomic, but the read and write are not one operation. Two concurrent CLI processes (e.g. two terminals running `testsprite setup` for different profiles, or `setup` + `auth remove`) can each read the same snapshot; the last rename wins and **silently drops** the other profile update - real credentials data loss.

## The fix

- Wrap `writeProfile` / `deleteProfile` in `withCredentialsLock()`
- Exclusive lock file at `{credentialsPath}.lock` via `openSync(..., 'wx')`
- Stale-lock reclamation when holder pid is dead or lock age > 30s
- Bounded retry (25ms back-off, 10s max wait)

## Tests

- Subprocess regression: parallel writers for `dev` + `staging` profiles - both survive
- Lock file is removed after `writeProfile` / `deleteProfile` complete

## Notes

- Does not overlap with open PRs (#4 Windows tests, #8 config/bundle, #26 failure-get preflight)
- Full `npm test` on Windows still has ~15 pre-existing failures covered by PR #4; new credentials tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential updates to remain reliable when multiple operations occur at the same time.
  * Ensured configuration, logout, and related flows report results only after credential persistence changes finish.

* **Tests**
  * Expanded coverage for concurrent credential updates, lock/cleanup behavior, and asynchronous command behavior.
  * Streamlined test setup by building once per run (global setup) and updated help snapshots and TTY expectations accordingly.

Solving Issue #109


jangjos-128:
Hi @SahilRakhaiya05 👋 — this is the recovery of your PR #32 , which was auto-closed by an error in our 2026-07-09 release process. Your commits are intact — thank you for the contribution!

It currently shows merge conflicts because `main` has moved on since your branch was cut. To clear them, update **your PR branch** (`BRANCH`) against the latest `main` — either option works and both update this PR in place:

  **Option A — merge (no force-push):**
  ```bash
  git checkout BRANCH
  git fetch upstream          # your remote for TestSprite/testsprite-cli (may be "origin")
  git merge upstream/main     # resolve conflicts, commit
  git push
```

Option B — rebase (cleaner history, needs force):
  ```bash
  git checkout BRANCH
  git fetch upstream
  git rebase upstream/main    # resolve conflicts
  git push --force-with-lease
```

 Once you push, the conflicts here resolve automatically and we'll take it into review.

 Sorry again for the extra step!

zeshi-du (changes requested):
Thanks for carrying this recovery forward — the lock design itself was already judged sound. Two things block it in its current shape:

1. The 07-14 conflict-resolution merge reintroduced `fileParallelism: false` in vitest.config (its own commit message says 'keep globalSetup + hermetic setupFiles + fileParallelism'). That's the exact condition from the original review: the build belongs in globalSetup (which you have), and the suite must keep file-level parallelism — serializing ~1.9k tests is the thing we asked to avoid.
2. There's in-flight maintainer work (landing shortly) that rebuilds the same vitest infra — build-once globalSetup + re-enabled file parallelism to fix Windows CI worker timeouts — touching the same three files this PR touches. To avoid double rework: once that lands on main, please rebase and drop this PR's vitest-infra changes entirely, keeping only the credentials cross-process lock + its tests.

Also note Lint & Format is genuinely red on the current head. The lock feature itself is still wanted!

zeshi-du:
Closing as **superseded by #272**, which just merged and fixes the same #77 race.

Head-to-head, #272 won on footprint and safety: it keeps `writeProfile`/`deleteProfile` synchronous (this PR's async conversion touches 11 files), leaves `vitest.config.ts` alone (the `fileParallelism: false` reintroduction from the 07-17 review was still unresolved here), and its stale-lock reclaim only fires on a *proven-dead* holder (`ESRCH`), where this branch's `catch { return true }` would also reclaim on e.g. `EPERM`.

**What this PR has that #272 doesn't — and that we still want:** the real two-process concurrency harness (`test/helpers/credentials-write-child.mjs` + `test/cli.subprocess.test.ts`). @jangjos-128 @SahilRakhaiya05 — if either of you rebases just that harness on top of current main as a small follow-up PR, we'll take it with full credit. That's the one missing piece of the merged fix.

Thanks for the recovery work and the original #32 design — the lockfile + stale-reclaim approach it pioneered is exactly what shipped.

PR #246, opened 2026-07-13
Recovered: fix(bundle): atomic re-commit via per-entry aside (#196 by @SahilRakhaiya05)

jangjos-128:
## Summary

Re-running `test failure get --out <dir>` can delete `meta.json` mid-commit; a failed re-commit then leaves the directory without a usable bundle.

This redesigns `commitBundle` to be **compatible with #162** (`isBundleOwnedEntry`): `--out` may contain the user's unrelated files, so a whole-directory swap is not safe.

## Approach

1. Stage the complete bundle in `<dir>/.tmp/` (unchanged)
2. Move only **bundle-owned** entries aside to a sibling `.<dir>.aside.<uuid>/` directory
3. Install new artifacts from `.tmp/`; `meta.json` lands last (completion signal)
4. On failure, restore aside entries (rollback) - prior bundle survives
5. On success, best-effort aside cleanup (cleanup failure does not roll back)

Foreign files (e.g. `notes.txt`, `src/`) are never touched.

## Testing

- `npx vitest run src/lib/bundle.commit.test.ts` - rollback, foreign-file preservation, aside cleanup
- `npx vitest run src/lib/bundle.test.ts -t "commit sweep"` - existing #162 ownership guard

Split from #8 per review feedback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bundle installation reliability by restoring the previous complete bundle if an installation step fails.
  * Preserved unrelated files during bundle updates.
  * Ensured cleanup failures after a successful update do not prevent the new bundle from being used.
  * Improved handling of incomplete or interrupted bundle updates.

### DIFF
--- a/src/lib/bundle.commit.test.ts
+++ b/src/lib/bundle.commit.test.ts
@@ -1,0 +1,124 @@
+import type * as NodeFsPromises from 'node:fs/promises';
+import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readdirSync,
+  readFileSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const renameMock = vi.hoisted(() => vi.fn());
+const rmMock = vi.hoisted(() => vi.fn());
+
+vi.mock('node:fs/promises', async importOriginal => {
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Verified byte-identical to the original refs/pull/196/head, and the request-changes condition is satisfied: meta.json is asided first (completion-signal ordering), with the skip in the sweep loop. Compatible with the #162 isBundleOwnedEntry contract. Thanks @SahilRakhaiya05!

PR #248, opened 2026-07-13
Recovered: fix: bound pagination and guard password file reads (#61 by @merlinsantiago982-cmd)

### LINKED ISSUE #58: Unbounded Pagination and Arbitrary File Read via --password-file

jangjos-128:
## Summary
- stop auto-pagination when the server repeats a cursor or exceeds a hard page budget
- guard `--password-file` reads with stat-first validation, regular-file checks, and a 64 KiB size cap
- add tests for cursor cycles, runaway pagination, valid password files, directory rejection, and oversized password files

## Why
A malicious or broken server could keep returning non-null cursors indefinitely, causing the CLI to loop and grow memory. `--password-file` also read arbitrary paths without checking the target type or size before sending the content as the project password.

## Testing
- npm test -- src/lib/pagination.test.ts src/commands/project.test.ts
- npm run typecheck
- npm run format:check
- npm run lint
- npm run build

Fixes #58

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `--password-file` validation for project creation and updates.
  * Rejects missing, unreadable, directory-based, or oversized password files before requests are sent.
  * Trims surrounding whitespace from password-file contents.
  * Prevents pagination from looping indefinitely when cursors repeat.
  * Stops automatic pagination after 1,000 pages and reports a clear error.

### DIFF
--- a/src/commands/project.test.ts
+++ b/src/commands/project.test.ts
@@ -7,6 +7,7 @@ import {
   type CliProject,
   type CliUpdateProjectResponse,
   createProjectCommand,
+  MAX_PASSWORD_FILE_BYTES,
   runCreate,
   runGet,
   runList,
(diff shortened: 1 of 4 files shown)


jangjos-128:
Hi @merlinsantiago982-cmd 👋 — this is the recovery of your PR #61 , which was auto-closed by an error in our 2026-07-09 release process. Your commits are intact — thank you for the contribution!

It currently shows merge conflicts because `main` has moved on since your branch was cut. To clear them, update **your PR branch** (`BRANCH`) against the latest `main` — either option works and both update this PR in place:

  **Option A — merge (no force-push):**
  ```bash
  git checkout BRANCH
  git fetch upstream          # your remote for TestSprite/testsprite-cli (may be "origin")
  git merge upstream/main     # resolve conflicts, commit
  git push
```

Option B — rebase (cleaner history, needs force):
  ```bash
  git checkout BRANCH
  git fetch upstream
  git rebase upstream/main    # resolve conflicts
  git push --force-with-lease
```

 Once you push, the conflicts here resolve automatically and we'll take it into review.

 Sorry again for the extra step!

zeshi-du (changes requested):
Re-reviewed the recovered diff in full. Good news first: the earlier data-loss concern doesn't apply to this final version — both pagination safety nets (max-pages cap, repeated-nextToken detection) throw an explicit UNAVAILABLE error rather than silently breaking with partial data. Nice.

What still stands is the original process ask: this bundles two independent concerns (pagination hardening + password-file read guards) that need separate review and separate revert paths. Please split it:
- **PR A — password-file guards**: we'll fast-track this half (it's a confirmed open gap: the ENOENT crash paths in project.ts are still live on main today, tracked in #79).
- **PR B — pagination safety caps**: reviewable on its own merits.

Both need a rebase onto current main regardless (this branch is conflicting after v0.4.0). Thanks @merlinsantiago982-cmd!

zeshi-du:
Closing this — most of it is now covered by other merged work, and I'd rather say so than leave it hanging for another month.

- **Pagination-safety half: already shipped.** `src/lib/pagination.ts` has `MAX_AUTO_PAGES = 1000`, repeated-cursor detection, and typed `max_pages_exceeded` / `repeated_next_token` errors — that landed via #48 back on 07-05.
- **`--password-file` guard half: superseded by #302**, merged today, which factors the guard into a shared `src/lib/secret-file.ts::readSecretFileGuarded()` covering the secret-file flags uniformly (typed `VALIDATION_ERROR`, exit 5, directory rejection, BOM stripping).

Credit to @merlinsantiago982-cmd for the original (#61) and to @jangjos-128 for recovering it after the July 9 force-push. Nothing here was wasted — it arrived alongside two other fixes covering the same ground, which is a coordination failure on our side, not a quality problem with the work.

Note for anyone landing on this from #58: that issue bundles two unrelated bugs. The pagination half is done; the remaining question is a policy one about size caps on secret-file reads, and I've recorded it there.

PR #213, opened 2026-07-07
fix(doctor): validate --output through the shared output-mode validator

Yazan-O:
## Summary

`doctor` resolved the global `--output` flag directly (`globals.output ?? 'text'`), bypassing the shared validator every other command uses — so an invalid mode (e.g. `--output yaml`) silently fell through instead of producing the standard typed `VALIDATION_ERROR`.

This routes doctor's option resolution through `resolveOutputMode` and adds command-wiring coverage: invalid `--output` now yields the standard `VALIDATION_ERROR`; `text` and `json` still work.

4 lines of src change + tests.

## Tests

- `vitest run src/commands/doctor.test.ts` — 14/14 pass
- `npm run lint:fix`, `npm run typecheck`, `npm run build` — pass
- Full `npm test` on this Windows host is blocked by the POSIX path/mode assumptions fixed in #207 — the two PRs compose.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of the doctor command’s output format so invalid values are rejected with a clear validation error.
  * Ensured valid output modes continue to work correctly, including both text and JSON results.

* **Tests**
  * Added coverage for command parsing and output-mode validation.
  * Verified successful report output in both text and JSON formats.

PR #207, opened 2026-07-07
test: make the Windows CLI test harness portable

Yazan-O:
## Summary

Running `npm test` on a Windows workstation currently fails in several suites, which blocks the documented contributor loop (CONTRIBUTING.md) for Windows users. This PR makes the test harness portable without changing any CLI behavior — the diff is tests + one test helper only.

- Make path- and CRLF-sensitive assertions separator/line-ending neutral.
- Make POSIX permission-bit assertions (`0o600`) POSIX-only — Node on Windows cannot represent them, so they asserted `0o666` and failed.
- Isolate subprocess tests from the real user profile by setting both `HOME` and `USERPROFILE`, and remove inherited `TESTSPRITE_*` env vars case-insensitively.
- Run subprocess build setup through the active npm entrypoint (with a Windows `.cmd` fallback) instead of assuming `npm` is directly spawnable.

## Why the isolation fix matters beyond CI

While validating this patch we hit the failure mode in the wild: running the unpatched suite on a Windows machine with a real TestSprite profile **overwrote `~/.testsprite/credentials` with the test fixture** (`sk-keep-me` @ `http://127.0.0.1:14979`), silently breaking every subsequent real CLI call with `UNAVAILABLE: fetch failed` until the file was restored by hand. The `HOME`/`USERPROFILE` isolation in this PR prevents that.

## Tests

- `npm test` — 50 files / 1846 tests pass on Windows 11 (Node 23) and under `npx -p node@22`.
- `npm run lint:fix`, `npm run typecheck`, `npm run build` — all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path and line-ending handling so CLI-related checks work more reliably on Windows and other environments.
  * Made file-path and profile checks more consistent by validating exact resolved paths instead of partial string matches.
  * Adjusted mode-related checks to avoid false failures on Windows while keeping restrictive permissions verified where applicable.

* **Tests**
  * Updated automated test setup to run builds and child processes more consistently across platforms.

PR #251, opened 2026-07-13
Recovered: fix(doctor): validate --output through the shared output-mode validator (#213 by @Yazan-O)

jangjos-128:
## Summary

`doctor` resolved the global `--output` flag directly (`globals.output ?? 'text'`), bypassing the shared validator every other command uses — so an invalid mode (e.g. `--output yaml`) silently fell through instead of producing the standard typed `VALIDATION_ERROR`.

This routes doctor's option resolution through `resolveOutputMode` and adds command-wiring coverage: invalid `--output` now yields the standard `VALIDATION_ERROR`; `text` and `json` still work.

4 lines of src change + tests.

## Tests

- `vitest run src/commands/doctor.test.ts` — 14/14 pass
- `npm run lint:fix`, `npm run typecheck`, `npm run build` — pass
- Full `npm test` on this Windows host is blocked by the POSIX path/mode assumptions fixed in #207 — the two PRs compose.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of the doctor command’s output format so invalid values are rejected with a clear validation error.
  * Ensured valid output modes continue to work correctly, including both text and JSON results.

* **Tests**
  * Added coverage for command parsing and output-mode validation.
  * Verified successful report output in both text and JSON formats.

### DIFF
--- a/src/commands/doctor.test.ts
+++ b/src/commands/doctor.test.ts
@@ -9,8 +9,9 @@
 import { mkdtempSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+import { Command } from 'commander';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import { CLIError } from '../lib/errors.js';
+import { ApiError, CLIError } from '../lib/errors.js';
 import { writeProfile } from '../lib/credentials.js';
 import type { DoctorDeps, DoctorReport } from './doctor.js';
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Verified against the original refs/pull/213/head — byte-identical recovery, and the --output validation now routes through resolveOutputMode with regression tests. Thanks @Yazan-O, and sorry again for the incident-closure detour.

PR #253, opened 2026-07-16
fix(auth): tighten Windows credentials ACL (#104)

### LINKED ISSUE #104: [Hackathon] Credential file is not owner-only on Windows (0o600 is a no-op there)

interferon0

writeCredentialsAtomic sets POSIX mode 0o600 and ensureRestrictiveMode chmods to 0o600 if over-permissive, but on Windows Node's fs mode only toggles the read-only bit, so the API key at %USERPROFILE%\.testsprite\credentials is not owner-only and can be readable by other local accounts on a shared/enterprise host. Worse, ensureRestrictiveMode's check reads a synthesized 0o666, calls chmodSync(0o600), "succeeds", and reports nothing, so operators believe the key is protected when it is not. Tighten the ACL on Windows, or at minimum warn.

src/lib/credentials.ts:182-198. On process.platform === 'win32', replace the POSIX chmod with an ACL tightening via icacls "<path>" /inheritance:r /grant:r "%USERNAME%:F" (spawn with an args array, never a shell string, to avoid path injection), OR at minimum emit a one-line stderr warning at write time that file permissions are not enforced on Windows. Do not leave the current silent "chmod succeeded" path. No new dependency. Tests: on win32 the ACL command runs / the warning is emitted; the POSIX path is unchanged. PR to follow once assigned.

### Confirmations

naufalfx805-source:
## What does this PR do?

Closes #104.

This stops the Windows credentials path from silently relying on POSIX `chmod(0o600)`, which Node cannot enforce on Windows ACLs.

Changes:

- On `win32`, `ensureRestrictiveMode()` now runs `icacls` with an args array: `icacls <path> /inheritance:r /grant:r <USERNAME>:F`.
- The POSIX chmod path remains unchanged for non-Windows hosts.
- If Windows ACL tightening cannot run, the CLI emits an explicit warning instead of silently pretending chmod protected the file.
- Adds unit coverage for the Windows ACL path using injected platform/spawn dependencies, so CI can verify this from any OS.

## Verification

- `npm test -- src/lib/credentials.test.ts` -> 24 passed, 1 skipped
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential-file permission tightening across platforms.
  * Added Windows support to apply restrictive ACLs to the current user when creating credentials.
  * Automatically warns when restrictive permissions cannot be applied (e.g., user lookup or permission command failure).
  * Kept the existing secure permission behavior for Unix-like systems.
* **Tests**
  * Added Windows-specific coverage for successful ACL tightening and warning scenarios.

### DIFF
--- a/src/lib/credentials.test.ts
+++ b/src/lib/credentials.test.ts
@@ -1,7 +1,7 @@
 import { mkdtempSync, statSync, readFileSync, writeFileSync, existsSync, mkdirSync } from 'node:fs';
 import { homedir, tmpdir } from 'node:os';
 import { join } from 'node:path';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import {
   DEFAULT_PROFILE,
   assertValidProfileName,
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Real gap confirmed (chmod 0o600 is a no-op for Windows ACLs) — this was on our internal fix list and your icacls approach is the right one: spawnSync with an argv array and shell:false (injection-safe), graceful degradation when USERNAME is missing or icacls fails, and DI-based tests that exercise the win32 path cross-platform. One non-blocking follow-up we may do later: DOMAIN\username disambiguation on domain-joined machines. Thanks @naufalfx805-source!

PR #256, opened 2026-07-16
fix(http): avoid retrying keyless writes (#103)

### LINKED ISSUE #103: [Hackathon] Transport retries replay non-idempotent writes with no idempotency key

interferon0

The HTTP client's transport-retry paths (network error, and bare 408/502/504) fire for ANY method, but the plain post/put/patch/delete helpers do not send an Idempotency-Key. So a response lost after the server committed it (a 502 from the edge on the response leg, or a socket reset) triggers a blind replay that can create a duplicate project, a duplicate test, or a second /tests/batch creation, costing credits and polluting the project. Gate the transport retries so a write without an idempotency key is not blindly replayed.

The retries are at src/lib/http.ts:467 (network error) and :530 (408/502/504); the dedicated trigger helpers DO send idempotency-key (http.ts:241, 269, 292, 311, 331) but the generic mutation helpers do not (call sites: project.ts:233, test.ts:803/2120/2428, delete at 1387/1634). Minimal fix: only auto-retry when method === 'GET' || method === 'HEAD' OR an idempotency-key header is present. Cleaner long-term: have the generic post/put/patch/delete mint an idempotency key by default. No new dependency. Tests: a transport error on a keyless POST is not retried; a GET or a keyed write still retries. PR to follow once assigned.

### Confirmations

naufalfx805-source:
## What does this PR do?

Closes #103.

This gates transport-level retries so non-idempotent writes without an `Idempotency-Key` are not blindly replayed after a network error or bare edge 408/502/504 response.

Changes:

- Allows transport retries for `GET`/`HEAD` requests.
- Allows transport retries for writes only when an `Idempotency-Key` header is present, case-insensitively.
- Leaves API-envelope retry behavior unchanged.
- Adds regression tests for keyless POST vs idempotency-key POST on both thrown transport errors and bare 502 edge responses.

## Verification

- `npm test -- src/lib/http.test.ts` -> 45 passed
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry handling for network failures and transient gateway responses.
  * Prevented automatic retries for non-idempotent requests unless an idempotency key is provided.
  * Continued supporting retries for safe requests and idempotent-keyed writes.

### DIFF
--- a/src/lib/http.test.ts
+++ b/src/lib/http.test.ts
@@ -260,6 +260,31 @@ describe('HttpClient error mapping', () => {
     expect(fetchImpl).toHaveBeenCalledTimes(4);
   });
 
+  it('does not retry transport errors for keyless writes', async () => {
+    const fetchImpl = vi.fn(async () => {
+      throw new Error('ECONNRESET');
+    });
+    const client = makeClient(fetchImpl as unknown as typeof fetch);
+    await expect(client.post('/projects', { body: { name: 'Checkout' } })).rejects.toBeInstanceOf(
+      TransportError,
+    );
+    expect(fetchImpl).toHaveBeenCalledTimes(1);
+  });
+
+  it('retries transport errors for writes with an idempotency key', async () => {
+    const fetchImpl = vi.fn(async () => {
+      throw new Error('ECONNRESET');
+    });
+    const client = makeClient(fetchImpl as unknown as typeof fetch);
+    await expect(
+      client.post('/projects', {
+        body: { name: 'Checkout' },
+        headers: { 'Idempotency-Key': 'op_123' },
+      }),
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Correct and minimal: transport-level retry now requires GET/HEAD or an Idempotency-Key header (case-insensitive), gated at both retry sites, with the 429/409 semantic retry path untouched. All standard write commands already attach idempotency keys, so the practical effect is scoped to keyless writes like cancelRun — which loses silent transport retry as an accepted behavior change (we'll consider giving cancel its own idempotency key as a follow-up). Thanks @naufalfx805-source!

PR #264, opened 2026-07-21
feat(test): GitHub-native CI output for "test run --all" (--gh-output, --summary-file)

### LINKED ISSUE #99: [Hackathon] Add "testsprite ci": one-shot run + gate + GitHub-native output

interferon0

A single command teams paste into one CI step: it runs a project's suite (wrapping the existing batch path), gates the job on the exit code, and when GITHUB_ACTIONS is set appends a Markdown results table to $GITHUB_STEP_SUMMARY, emits one ::error:: workflow-command line per failed test (title + dashboard link) so failures annotate the PR checks tab, and writes a stable machine summary to stdout or --summary-file. It auto-detects the runner and degrades to plain output off-CI. This collapses "run + gate + PR feedback + dashboard artifact" into one call, the exact shape gh, Cypress, and Playwright converged on.

The batch --all --wait path already computes {passed, failed, timedOut, total} plus per-test runId/dashboardUrl/error (src/commands/test.ts:5743), so ci is a presenter over data the CLI already assembles, with zero new wire calls. The backend already stamps source: 'github_action' (src/lib/runs.types.ts:249), so the provenance loop is pre-wired. New src/commands/ci.ts + registration in src/index.ts + a dry-run sample; reads GITHUB_STEP_SUMMARY/GITHUB_ACTIONS from process.env, writes via fs. No new dependency. Tests: summary/annotation/machine-summary formats; off-CI plain fallback. PR to follow once assigned.

### Confirmations

Andy00L:
Reshaped per the maintainer review: the top-level `ci` command is gone; the CI-native output layer now attaches to the existing `test run --all` path.

- New `src/lib/gh-output.ts` keeps the reducer (`summarizeAcceptedPayload`), the `$GITHUB_STEP_SUMMARY` Markdown renderer, and the `::error::` annotations emitter, with their unit tests moved to `src/lib/gh-output.test.ts`.
- `test run --all` gains `--gh-output` (force the GitHub-native surfaces even off-Actions) and `--summary-file <path>` (write the reduced `{total, passed, failed, timedOut, runs[]}` JSON artifact). Auto-enabled when `GITHUB_ACTIONS=true`, so the official Action gets all of it for free by invoking `test run --all --wait`.
- Emission happens after the batch envelope prints and before the gate throws, so artifacts land even when the batch exits non-zero; a broken summary/step file downgrades to a stderr warning and never masks the gate's exit code.
- `--output` contract defect fixed by construction: nothing extra is written to stdout (the envelope owns it; workflow commands are only emitted under Actions or explicit `--gh-output`); the summary file is a machine artifact written in both output modes.
- Both flags are `--all`-only and rejected loudly otherwise (same rule as `--filter`).
- Verified against current main: typecheck, build, full suite (2023 tests), lint and prettier clean.

Closes #99 CLI-side per the direction note (the one-shot entrypoint ships as the official GitHub Action).

Discord: interferon0

## Summary by CodeRabbit

* **New Features**
  * Added GitHub Actions output support for `testsprite test run --all --wait`, including workflow error annotations and a `$GITHUB_STEP_SUMMARY` job summary.
  * Introduced `--gh-output` and `--summary-file` (with `--summary-file` writing a reduced JSON summary).
* **Bug Fixes**
  * Improved robustness: malformed/unexpected payloads produce an empty summary without failing, and step-summary writing issues degrade to warnings.
* **Tests**
  * Added/extended coverage for summary reduction, step-summary rendering, and correct routing of annotations/output across modes.

### DIFF
--- a/src/commands/test.run.spec.ts
+++ b/src/commands/test.run.spec.ts
@@ -5,7 +5,7 @@
  * sleep injection is wired through `TestDeps.sleep` to avoid real delays.
  */
 
-import { mkdirSync, mkdtempSync, writeFileSync } from 'node:fs';
+import { mkdirSync, mkdtempSync, readFileSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import type { Command } from 'commander';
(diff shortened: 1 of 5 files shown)


zeshi-du:
Maintainer verdict: **the engineering is good, but a new top-level `ci` command is the one shape we decided not to build** — asking for a reshape rather than closing, because most of this PR survives it.

Context, restated from #99 (07-18): the CI close-out ships as an **official GitHub Action** (in progress internally), not a new CLI subcommand. A permanent top-level command is a long-term maintenance commitment (the concern raised in #99's original triage), and `testsprite ci` as a wrapper around `test run --all --wait` adds a second way to do the same thing — which we've been deliberate about avoiding.

**What we do want from this PR — the CI-native *output*, which is the genuinely missing half:**
- the `::error::` per-failed-run annotations emitter
- the `$GITHUB_STEP_SUMMARY` Markdown table renderer
- the `--summary-file` stable-JSON artifact

Reshape: attach those to the existing `test run` path (activated by `GITHUB_ACTIONS=true`, or an explicit `--gh-output` flag), drop `src/commands/ci.ts` as a command registration, keep the reducer/renderer/tests. The official Action then gets them for free by invoking `test run --all --wait`.

One defect to fix in the same pass: the JSON summary writes to stdout unconditionally, ignoring the global `--output text` flag (your own test at `ci.test.ts:140` pins the bypass). Under the reshape this becomes moot on stdout, but the summary-file path should still respect the flag's contract.

Happy to re-review fast — the test discipline here is exactly what we want to merge.

Andy00L:
Reshaped as requested: the top-level command is gone; the reducer/renderer/emitter moved to src/lib/gh-output.ts and hook into `test run --all` (auto on GITHUB_ACTIONS=true, forced by --gh-output). --summary-file is now a machine artifact on the run path and stdout stays owned by the batch envelope, which also settles the --output contract defect. Full suite green (2023 tests). Ready for the fast re-review.

Andy00L:
CI red root-caused: the three failing run-spec tests inherited the runner's real GITHUB_ACTIONS=true, which (correctly) activated the new gh-output layer and appended ::error:: lines to the stdout they JSON.parse. Fixed by injecting an empty env into those deps; verified locally under GITHUB_ACTIONS=true (full suite 2023 green).

Andy00L:
All three review points addressed:
1. `--gh-output` / `--summary-file` now require `--all --wait` (without `--wait` there is no terminal envelope to reduce; both flags reject loudly, help text updated).
2. Annotations no longer touch JSON stdout: `emitGithubOutputs` gained a dedicated `annotations` sink and the run path wires it to stderr under `--output json` (the Actions runner parses workflow commands on both streams); integration test asserts stdout stays parseable JSON while `::error::` lands on stderr.
3. `summarizeAcceptedPayload` now treats `null`/non-record payloads and null rows as no-data instead of crashing; regressions added for `null`, primitives, and `accepted: [null, 42, ...]`.
Full suite green under GITHUB_ACTIONS=true (2027 tests).

zeshi-du (approved):
Re-reviewed after the reshape — this is exactly what was asked: no new top-level command, `--gh-output`/`--summary-file` attached to `test run --all --wait` with GITHUB_ACTIONS auto-enable, artifact still written on a failing gate, stdout contract kept clean. Fast turnaround appreciated.

PR #265, opened 2026-07-21
feat(test): add "test export" / "test import" to round-trip test DEFINITIONS for version control

### LINKED ISSUE #125: [Hackathon] Add "test export" / "test import" to round-trip test DEFINITIONS for version control

interferon0

There is no way to get a test's definition out of (or back into) TestSprite as a file, so test definitions cannot be version-controlled, reviewed in PRs, backed up, or migrated — the tests-as-code model that made Checkly and Terraform sticky. Add test export <id> [--out <file>]: compose the existing GET /tests/{id} (metadata) + GET /tests/{id}/code into one deterministic <id>.testsprite.json (stable key order, a schemaVersion, and the codeVersion recorded as provenance). Add test import <file>: create (POST /tests) or update (PUT /tests/{id} + PUT .../code, honoring the existing codeVersion optimistic-concurrency guard) from that file. Note: this is DEFINITION export, distinct from the filed JUnit RESULTS export (#95) — different artifact, different purpose.

Honest limitation disclosed up front: backend tests round-trip losslessly (metadata + Python code), but a frontend test's authored planSteps[] is write-only on the wire (PUT /tests/{id}/plan-steps exists; there is no GET, and GET /tests/{id} returns only planStepCount) — so FE export carries metadata + generated code with an explicit planUnavailable: true marker instead of silently dropping the plan. If that gap matters to reviewers, v1 can be scoped to backend tests. Reuses the existing create/update/code-put request shapes and idempotency-key minting; file writes honor the existing --out safety guards. No new dependency. Natural follow-up (separate issue): a declarative sync <dir> reconcile. PR to follow once assigned.

### Confirmations

Andy00L:
Round-trip a test DEFINITION (metadata + code with `codeVersion` provenance) to a versionable JSON file, so definitions can be reviewed, backed up, and migrated.

- `test export <test-id>` writes the definition (`--out <file>` + `--force`, stdout by default). Frontend plans are write-only on the API, so FE exports carry `planUnavailable: true` and a stderr note.
- `test import <file>` creates or updates from a definition file: a `testId` in the file selects update (code PUT replays the recorded `codeVersion` as `If-Match`, so a drifted server copy fails loudly with the existing 412 contract); no `testId` creates.
- 5 unit tests (export shape, FE marker, create vs update paths, If-Match replay); `test` surface test and help snapshot updated for the two new subcommands.
- Verified against current main: typecheck, build, full suite (2007 tests), lint and prettier clean.

Closes #125

Discord: interferon0

## Summary by CodeRabbit

* **New Features**
  * Added `test export` to generate deterministic, versioned test definition JSON to stdout or `--out` (supports `--dry-run` and prevents overwrites unless `--force`).
  * Added `test import` to create or update tests from a definition file (validates schema fields, supports `--dry-run`, and reports `created`/`updated`).
  * Frontend exports set `planUnavailable` and emit a write-only warning.

* **Tests**
  * Expanded end-to-end coverage for export/import flows, schema/version validation, BOM-prefixed inputs, overwrite/force behavior, and idempotency-key handling.

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -31,6 +31,8 @@ import {
   runFailureGet,
   runFailureSummary,
   runGet,
+  runExport,
+  runImport,
   runLint,
   runList,
   runOpen,
@@ -128,9 +130,11 @@ describe('createTestCommand — surface', () => {
       'delete',
       'delete-batch',
       'diff',
+      'export',
       'failure',
       'flaky',
       'get',
+      'import',
       'lint',
       'list',
       'open',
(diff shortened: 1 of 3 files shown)


zeshi-du:
Maintainer review: **request changes** — solid concept, five concrete gaps, plus a scope note.

1. **BOM handling (Windows regression):** `runImport`'s `readFileSync` + `JSON.parse` doesn't go through `stripBom()` — every other JSON file-read in `test.ts` does, precisely because PowerShell 5.1's `Set-Content -Encoding utf8` writes a BOM (see the helper's own doc comment). As written, `test import` fails on exactly the platform CI tests for. Add `stripBom()` + a BOM-prefixed test file case.
2. **Idempotency parity:** the import create path mints `'idempotency-key': cli-import-create-${randomUUID()}` internally, with no `--idempotency-key` flag and no stderr echo. Every other mutating command in this file (`create`, `create-batch`, `update`, `delete`, `code put`, `plan put`) exposes the flag + auto-mint + stderr echo, for the reason documented at the `test create` site: a transport-level retry after an ambiguous failure must be able to reuse the same key, or it duplicates the test. `runImport` needs the same triad on both the create POST and the code PUT.
3. **`codeVersion: null` on the update path silently omits `If-Match`** → unconditional overwrite with no warning. `code put` handles this exact case with `If-Match: '*'` plus a stderr notice ("auto-fetched codeVersion=null on legacy row..."). Match that, and add a test — this branch currently has zero coverage.
4. **`--out`/`--force` file-write path is untested:** all five new tests exercise stdout export only; the overwrite guard and the actual file write are the safety-critical surface here.
5. **Error conflation:** a typo'd path (ENOENT) and malformed JSON produce the same "not readable valid JSON" message. Split them — "file not found: <path>" vs the JSON error.

**Scope note:** issue #125 was triaged *deferred* on 07-03 (schema-freeze vs in-flight server-side test-entity work) and that hold hasn't formally lifted — we're re-evaluating it now that the wire contract is stability-committed, and will record the outcome on #125. Fixing the five items above is worthwhile regardless of timing; landing waits on that call.

Andy00L:
All five points addressed:
1. `runImport` now reads through `stripBom()` + a BOM-prefixed test case.
2. Idempotency triad on both mutations: `--idempotency-key` flag on `test import`, auto-mint `cli-import-<uuid>`, stderr echo (create-site condition); per-mutation keys derive deterministically (`:meta`/`:code`/`:create`) so a replayed key reuses the same wire keys.
3. `codeVersion: null` now sends `If-Match: *` with a stderr notice (mirrors `code put`'s legacy-row handling) + dedicated test.
4. `--out`/`--force` file-write path covered: write, overwrite-guard rejection, forced overwrite.
5. ENOENT and malformed JSON are split into `file not found: <path>` vs `is not valid JSON: <path>`.
Full suite green (2012 tests). Understood on the scope hold — ready whenever the #125 re-evaluation lands.

Andy00L:
Lint red fixed: the BOM fixture sat in a template literal, which no-irregular-whitespace rejects (plain strings are exempt, which is why the existing BOM fixtures pass). Moved to a concatenated string; eslint and the suite are green.

Andy00L:
Adopted the suggested \uFEFF escape form for the BOM fixture (the interim concatenated-string fix already unblocked lint; this matches the proposed diff exactly).

zeshi-du:
Re-reviewed properly. **All five defects from the July review are fixed** — BOM stripping on import, the idempotency key on the write path, `codeVersion` no longer dropping `If-Match`, `--out` covered by tests, and ENOENT distinguished from malformed JSON. That's a complete response to review feedback and it deserved a faster acknowledgement than 19 days.

What's blocking is on me, not the code. #125 was deferred on 2026-07-03 because an export format would freeze a test-entity shape we were actively changing; I promised on 07-23 to record whether that deferral was lifted, and never did. I'm posting the analysis on #125 now and will land the decision there rather than leaving it implicit here.

Two practical notes meanwhile:
- The branch has gone `CONFLICTING` — `main` has moved two releases (v0.5.0 on 08-06, v0.6.0 on 08-12). A rebase now means we can merge the moment the #125 call lands.
- Worth re-checking the import path against the shared `readSecretFileGuarded()` helper that just landed in `src/lib/secret-file.ts` (#302) — if `test import` reads a file by user-supplied path, it should reuse that guard rather than its own try/catch, so the `VALIDATION_ERROR`/exit-5 contract is identical across every file flag.

zeshi-du:
Thanks for the patience — this sat 19 days after you addressed CodeRabbit's review, and that silence is on us. Now that #125's deferral is lifted, here's what's needed to land it.

**1. The conflict is a false alarm, not competing logic.**

Both this branch and `main` insert new content at the exact same seam in `src/commands/test.ts`: right after `CliLintReport`'s closing brace, a blank line, and the opening `/**` of the next doc comment. After this branch was cut, `main` added a `toLintIssue()` helper (with its own doc comment) at that seam; this branch adds `CliTestDefinition` / `ExportOptions` / `runExport` / `ImportOptions` / `runImport` at the identical point. Neither side touches or deletes a line the other wrote — two independent insertions competing for one anchor.

Resolution: rebase onto current `main` and keep both blocks. Order between them doesn't matter — they're self-contained top-level declarations with no cross-references. No code changes on either side, just picking an order when Git asks.

**2. Missing `CHANGELOG.md` entry.** Please add a bullet under `## [Unreleased]` / `### Added` for `test export` and `test import`.

**3. Non-atomicity in `runImport`'s update path — flagging, not blocking.** The metadata PUT (`:meta`) and the code PUT (`:code`, with `If-Match`) are sequential and independent, with nothing to roll back the metadata write if the code PUT fails. Two different failure shapes here, and they shouldn't be conflated: a transport-level failure between the two PUTs is safely retryable, because re-running with the same idempotency key reuses the same derived `:meta` / `:code` keys rather than minting new ones. A 412 on the code PUT (the server's `codeVersion` moved) is a different, deterministic case, and it's *supposed* to fail loudly rather than be retried past — that part is working as designed.

I don't think this needs to hold up the PR. A doc-comment note on `runImport` about the partial-state window (new metadata, stale code, until re-run) would be a good follow-up.

Separately: see the note on #125 about whether the import path's file read can reuse `readSecretFileGuarded()` — tracked there, not repeating it here.

zeshi-du (approved):
Approving the direction.

**All five defects from the July 23 review are fixed** — checked each against the current diff, not just your word for it:

1. **BOM handling** — `runImport` now does `stripBom(readFileSync(absolute, 'utf8'))` before `JSON.parse`, matching every other JSON read in this file, plus a BOM-prefixed test case.
2. **Idempotency parity** — `--idempotency-key` flag, auto-mint with a stderr echo, and `:meta` / `:code` keys derived from one base key so a retry after an ambiguous failure replays the same wire keys instead of minting new ones.
3. **`codeVersion: null`** no longer drops `If-Match` silently — it now sends `If-Match: *` with a stderr notice, matching `code put`'s handling of a legacy row, with a test covering it.
4. **`--out` / `--force`** is covered: "export --out writes the file, refuses to overwrite without --force, overwrites with it."
5. **ENOENT vs malformed JSON** are two distinct messages now ("file not found: …" vs "is not valid JSON: …").

That's a complete response to review feedback.

**The conflict is one real spot, and it's an adjacent insertion, not competing logic.** Both branches insert new top-level code at the identical seam in `src/commands/test.ts`, right before the pre-existing `test lint` doc comment: `main` added a `toLintIssue()` helper (with its own doc comment) there after this branch was cut; you add `CliTestDefinition` / `ExportOptions` / `runExport` / `ImportOptions` / `runImport` at the same point. Neither side touches a line the other wrote. Resolution: keep both blocks; the order between them doesn't matter, since they're self-contained top-level declarations with no cross-references. `src/commands/test.test.ts` and `test/__snapshots__/help.snapshot.test.ts.snap` both changed on both sides too, but auto-merge without conflict markers — run the snapshot test after rebasing to confirm the merged help text is exact.

**Rebase**

```
git fetch upstream main
git rebase upstream/main
# only src/commands/test.ts should show conflict markers — keep both inserted
# blocks (main's toLintIssue() and your CliTestDefinition/runExport/runImport),
# in either order
git add src/commands/test.ts
git rebase --continue
npm run lint && npm run typecheck && npm test
git push --force-with-lease
```

CI is green, but that run is from 2026-07-23 — before both v0.5.0 (08-05) and v0.6.0 (08-12) — so treat it as informative, not current; the post-rebase run is the real gate.

**Scope, as a condition of merge, not a formality** — this is a file format we own forever once it ships, so before I merge, confirm in this thread which of these you've verified:

1. **The export is a projection of the API's existing shapes, not a new schema with its own versioning story.** Right now `def.schemaVersion !== 1` gates `import`, and `schemaVersion: 1` is stamped on every `export` — that's a fixed compatibility check today, but it's also exactly the shape of the `formatVersion` field I flagged as the signal we're inventing a format. Tell me which one it is, and why — on its face it reads like the latter, not the former.
2. Round-tripping is lossless or explicitly lossy, never silently lossy.
3. This is single-definition version control, not a suite/grouping concept — Test Lists own that, and export/import shouldn't grow toward it.

One item stays on #125 rather than repeating here in full: whether `runImport`'s file read can reuse `readSecretFileGuarded()` (`src/lib/secret-file.ts:40`). One correction to that ask, since it changes what "reuse" means: that helper's `.trim()` already strips a leading BOM (`U+FEFF` is ECMAScript whitespace — its own doc comment says so), so adopting it would make your separate `stripBom()` call redundant, not something you'd still need on top of it. The stat → ENOENT/EACCES/EISDIR → `VALIDATION_ERROR` mapping it does is still the same contract you hand-rolled for defect 5.

Missing before merge: a `CHANGELOG.md` entry under `## [Unreleased]` for `test export` / `test import`.

zeshi-du (changes requested):
I need to withdraw the approval I gave on 2026-08-19, and explain properly why, because you have turned around every round of review we've asked for and this is our error rather than yours.

**What went wrong on our side.** That approval was given on the basis that the five defects from the July review were fixed and CI was green — both true. It was *not* given against the three merge conditions we wrote down internally when we lifted the deferral on this feature, which is the check that should have happened before an approval, not after. Having now done it properly, two of the three are not met.

**1. The export must be a projection of shapes the API already returns.**

The export carries `schemaVersion: 1`, and the import side has a test that rejects a wrong `schemaVersion`. That is us inventing and versioning a file format, which is the specific outcome that condition exists to prevent — once a `schemaVersion: 1` ships, we owe the world a migration story for `2`, forever, for a convenience feature.

**2. Round-trip must be lossless, or explicitly lossy — never silently lossy.**

Two places break this, and the first is the one that actually worries me:

- **`GET /tests/{id}/code` returns a presigned URL rather than a body once the code exceeds roughly 100KB.** `test code get` dereferences that URL before showing it to the user. This PR stores the response verbatim as the test's source. So for any test above that threshold, the exported file contains a presigned URL where the source should be — and on import, once that URL has expired, the test's source *becomes* that dead URL string. **Large tests are silently corrupted, and the corruption only becomes visible later, on a run.** `isPresignedCodeUrl` appears zero times in the diff, so there is no branch that could be catching this.
- The export drops `priority`, `produces`, `consumes` and `category`, while the test that covers it is named "backend: lossless".

**What unblocks a merge**, concretely:

1. Drop `schemaVersion` and project the existing wire envelope instead.
2. Detect a presigned code URL and either dereference it into bounded source, or refuse to export that test with a clear error naming the test. Silently writing the URL is the one option that isn't acceptable.
3. Either carry `priority` / `produces` / `consumes` / `category`, or declare the loss — in the exported file itself and in `--help` — and rename the test so it doesn't claim losslessness it doesn't have.

Separately: `main` has shipped four releases since this branch was cut (v0.8.0 → v0.11.0), so `src/commands/test.ts` now conflicts. Rebase whenever you pick this back up — no rush, and there's no need to touch it until you've decided whether you want to take the three points above.

I'm sorry this took four weeks to say. The feature is one we want; the shape isn't right yet.

zeshi-du (changes requested):
Hi @Andy00L — round-tripping test definitions through a file is genuinely useful, and the test coverage here is thorough. A few things to settle before merge:

- `schemaVersion: 1` introduces a new file-format contract — let's agree on the versioning/compatibility guarantees before it ships, so we're not locked into an undocumented shape.
- For large tests, `test export` can persist a presigned URL instead of the actual source code, which risks silent data loss once the URL expires. Please have `export` always write the real source, or clearly flag when it hasn't.
- `priority`, `produces`, `consumes`, and `category` are dropped on export with no warning, unlike the existing `planUnavailable` disclosure for FE plans — please add a similar explicit warning so round-tripping doesn't quietly lose fields.
- Please rebase `src/commands/test.ts` against `main`.

PR #266, opened 2026-07-21
feat(http): validate API responses at runtime with valibot instead of blind casts

### LINKED ISSUE #102: [Hackathon] Validate API responses at runtime with valibot instead of blind casts

interferon0

Every successful response body is a compile-time-only cast (as T) with zero runtime validation, so a partial or drifted server response silently poisons the CLI's "reliable typed output" contract: the agent reasons over undefined, or the process dies with an opaque TypeError it cannot map to an exit code. Add valibot schemas for the wire types and validate at the single choke point in the HTTP client, converting drift into a typed INTERNAL/UNAVAILABLE envelope (with requestId) instead of undefined output or a crash. This finally makes valibot earn its place and gives agents guaranteed-shape output.

The blind cast is at src/lib/http.ts:497 (return { body: (await response.json()) as T, ... }); the wire types are plain interfaces in src/lib/runs.types.ts (erased at runtime). Add src/lib/response-schemas.ts with valibot schemas (RunResponse, TriggerRunResponse, RerunResponse, ListRunsResponse, CliLatestResult, ...) and validate in HttpClient.request. valibot is ALREADY a declared runtime dependency (used only for input validation today, src/lib/validate.ts), so this is the intended-but-unrealized use its own header notes: no new dependency. Tests: a drifted/partial body yields a typed error envelope with requestId, not undefined/crash; a valid body passes untouched. PR to follow once assigned.

### Confirmations

Andy00L:
Validate API responses on the run paths against valibot schemas at the HTTP client, instead of blind `as T` casts.

- New `src/lib/response-schemas.ts`: `looseObject` schemas for the run-path payloads (trigger, rerun, batch rerun, batch fresh, get run, list runs); unknown extra keys are preserved so additive server changes stay non-breaking.
- `RequestOptions` gains an optional `schema` field; when present, `requestWithMeta` runs `v.safeParse` on the OK-path JSON. Shape drift now surfaces as a typed `INTERNAL` envelope naming the request path and the first mismatched field paths (never echoing the body), instead of `undefined` output or an opaque crash downstream.
- Wired by the typed run helpers only (`triggerRun`, `triggerRunWithMeta`, `triggerRerun`, `triggerBatchRerun`, `triggerBatchRunFresh`, `getRun`, `listTestRuns`); generic `get`/`post`/... callers are unchanged and stay opt-in.
- Uses the `valibot` dependency already present in the project; no new dependency.
- Verified against current main: typecheck, build, full suite, lint and prettier clean.

Closes #102

Discord: interferon0

## Summary by CodeRabbit

* **New Features**
  * Added response validation for run-related API requests.
  * Supports typed schemas for HTTP requests and responses.
  * Preserves additional fields returned by the server for forward compatibility.
  * Provides detailed validation errors when response data does not match expectations.

* **Bug Fixes**
  * Prevents malformed API responses from being silently accepted as valid data.
  * Added coverage for schema validation, compatible response changes, and unvalidated requests.

### DIFF
--- a/src/lib/http.ts
+++ b/src/lib/http.ts
@@ -1,7 +1,16 @@
 import { randomUUID } from 'node:crypto';
+import * as v from 'valibot';
 import type { ErrorCode } from './errors.js';
 import { ApiError, InterruptError, RequestTimeoutError, TransportError } from './errors.js';
 import { VERSION } from '../version.js';
+import {
+  BATCH_RERUN_RESPONSE_SCHEMA,
+  BATCH_RUN_FRESH_RESPONSE_SCHEMA,
+  LIST_RUNS_RESPONSE_SCHEMA,
+  RERUN_RESPONSE_SCHEMA,
+  RUN_RESPONSE_SCHEMA,
+  TRIGGER_RUN_RESPONSE_SCHEMA,
+} from './response-schemas.js';
 import type {
   TriggerRunBody,
   TriggerRunResponse,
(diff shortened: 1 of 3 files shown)


zeshi-du:
Maintainer review: **accepted — merging.** And recording the design sign-off this was parked on:

The validation policy question that parked #102 on 07-03 ("schemas must be passthrough-tolerant so additive server changes never break the CLI") is **settled by this PR's rules**: `v.looseObject` everywhere (unknown fields pass through *and are preserved*), `openWireLiteral()` for enum-like fields (a new server-side status value degrades gracefully instead of hard-failing), `nullish`/`optional` threaded from documented fixture evidence, and fail-closed only on genuinely required fields with the wrong type. That is exactly the policy we wanted — treat it as the reference for any future schema additions.

Scope note for the record: this wires the 7 run-path shapes (the ones agents' exit-code decisions depend on) and deliberately leaves ~15 other response shapes on the blind cast; `ME_IDENTITY_SCHEMA` is exported but unwired, which is fine as a forward-looking stub. Follow-up issue for incremental coverage coming right after merge.

Thanks @Andy00L — strongest of your three this week.

zeshi-du (approved):
LGTM — see verdict comment.

PR #272, opened 2026-07-22
fix(auth): lock credential profile mutations

### LINKED ISSUE #77: [Hackathon] Concurrent Profile Writes Silently Drop Updates (No File Lock)

naufalfx805-source:
## Summary
- Wrap credential profile write/delete mutations in an advisory lock held across the full read-modify-write cycle.
- Reclaim stale credential lock files and release only the lock owned by the current mutation token.
- Keep missing-profile deletion lightweight without creating a credentials file.

Fixes #77

## Claim
- Issue #77 is assigned to @naufalfx805-source via /assign.

## Test plan
- npm test -- src/lib/credentials.test.ts
- npm test -- src/commands/auth.test.ts
- npm run typecheck
- npm run lint

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of credential updates when multiple operations occur concurrently.
  * Stale credential locks are now safely reclaimed, preventing updates from becoming stuck.
  * Deleting a missing profile now leaves no credentials file behind.

* **Tests**
  * Added coverage for stale-lock recovery and cleanup behavior.

### DIFF
--- a/src/lib/credentials.test.ts
+++ b/src/lib/credentials.test.ts
@@ -157,6 +157,19 @@ describe('writeProfile', () => {
     expect(file.dev).toEqual({ apiKey: 'sk-dev', apiUrl: 'https://dev' });
   });
 
+  it('reclaims stale mutation locks and removes the lock after writing', () => {
+    const lockPath = `${credentialsPath}.lock`;
+    writeFileSync(
+      lockPath,
+      `${JSON.stringify({ pid: process.pid, createdAt: Date.now() - 60_000, token: 'stale' })}\n`,
+    );
+
+    writeProfile('default', { apiKey: 'sk-new' }, { path: credentialsPath });
+
+    expect(readProfile('default', { path: credentialsPath })).toEqual({ apiKey: 'sk-new' });
+    expect(existsSync(lockPath)).toBe(false);
+  });
+
   it('does not leak the api key into the on-disk file format aside from the value itself', () => {
     writeProfile('default', { apiKey: 'sk-secret-12345' }, { path: credentialsPath });
     const onDisk = readFileSync(credentialsPath, 'utf-8');
@@ -168,6 +181,7 @@ describe('writeProfile', () => {
 describe('deleteProfile', () => {
   it('returns false when the profile is missing', () => {
     expect(deleteProfile('nope', { path: credentialsPath })).toBe(false);
+    expect(existsSync(credentialsPath)).toBe(false);
   });
 
   it('removes the named profile and leaves others intact', () => {
--- a/src/lib/credentials.ts
+++ b/src/lib/credentials.ts
@@ -5,6 +5,7 @@ import {
   readFileSync,
   renameSync,
   statSync,
+  unlinkSync,
   writeFileSync,
 } from 'node:fs';
 import { spawnSync, type SpawnSyncReturns } from 'node:child_process';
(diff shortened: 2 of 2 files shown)


zeshi-du:
Maintainer review: **accepted — merging.** This is the right shape for #77:

- `wx`-flag exclusive create is genuinely atomic, no new dependency
- stale reclaim (30s + liveness probe) defaults to *assume-alive* unless `ESRCH` is proven — the safe direction
- `assertHeld()` re-check before the final atomic write closes the stolen-lock race (nice catch turning the CodeRabbit note into a second commit)
- sync signatures preserved, so no caller churn; and no test-runner config changes

One follow-up we'd like (not blocking): the added test simulates a stale lock in-process, but nothing exercises a genuine cross-process race. A small follow-up PR spawning two child processes writing different profiles concurrently (the pattern in #244's `test/helpers/credentials-write-child.mjs` is a good reference) would lock this in for real. Feel free to claim it, or we'll file it as a good-first issue.

Thanks @naufalfx805-source — second precision fix this month.

zeshi-du (approved):
LGTM — see verdict comment.

naufalfx805-source:
I'll take the non-blocking follow-up for a real cross-process credentials mutation race test, using the #244 child-process helper pattern as reference. I'll open a focused PR for that next.

PR #273, opened 2026-07-22
fix(completion): cross-platform shell auto-detection for Windows paths and .exe extensions

### LINKED ISSUE #275: fix(completion): cross-platform shell auto-detection for Windows paths and .exe extensions
### Problem
When running `testsprite completion` without specifying a shell argument, `detectShell(env)` in `src/commands/completion.ts` inspects `$SHELL`.
On Windows (Git Bash, MSYS2, WSL), `$SHELL` uses backslashes (e.g. `C:\Program Files\Git\bin\bash.exe`) or `.exe` extensions.
Currently, `detectShell` uses `shellPath.slice(shellPath.lastIndexOf('/') + 1)`, which fails on backslashes and `.exe` extensions, throwing a `VALIDATION_ERROR` (exit 5).

### Proposed Fix
Enhance `detectShell` to split on `/[/\\]/`, strip trailing `.exe`, and lowercase before checking against supported shells (`bash`, `zsh`, `fish`).

sandman-sh:
Closes #275

## Summary

Fixes shell auto-detection in `testsprite completion` when invoked without an explicit shell argument.

### Problem
Currently, `detectShell(env)` in `src/commands/completion.ts` uses `shellPath.slice(shellPath.lastIndexOf('/') + 1)`.
On Windows or cross-platform environments using Git Bash / MSYS2 / WSL:
1. `$SHELL` uses backslashes (e.g. `C:\Program Files\Git\bin\bash.exe`). `lastIndexOf('/')` returns `-1`, resulting in the full path string `C:\Program Files\Git\bin\bash.exe` which fails `isShell()`.
2. `$SHELL` filenames end in `.exe` (e.g. `bash.exe`, `zsh.exe`, `fish.exe`), causing `isShell("bash.exe")` to return `false`.

### Solution
Enhances `detectShell(env)` to:
- Split on both forward `/` and backward `\` slashes (`/[/\\]/`).
- Lowercase and strip trailing `.exe` extensions (`.toLowerCase().replace(/\.exe$/, '')`).
- Match against `isShell()` (`bash`, `zsh`, `fish`).

### Test Coverage
Updated `src/commands/completion.test.ts` with unit tests for Windows paths, `.exe` suffixes, mixed slashes, and uppercase names. All 2,002 tests pass.

### DIFF
--- a/src/commands/completion.test.ts
+++ b/src/commands/completion.test.ts
@@ -23,8 +23,16 @@ describe('isShell / detectShell', () => {
     expect(detectShell({ SHELL: '/usr/local/bin/fish' })).toBe('fish');
   });
 
+  it('detects shells with Windows backslashes, .exe suffixes, and uppercase paths', () => {
+    expect(detectShell({ SHELL: 'C:\\Program Files\\Git\\bin\\bash.exe' })).toBe('bash');
+    expect(detectShell({ SHELL: 'C:\\tools\\zsh.exe' })).toBe('zsh');
+    expect(detectShell({ SHELL: '/usr/bin/FISH.EXE' })).toBe('fish');
+    expect(detectShell({ SHELL: 'C:/Program Files\\Git\\bin/bash.exe' })).toBe('bash');
+  });
+
   it('returns undefined for an unknown or missing shell', () => {
     expect(detectShell({ SHELL: '/bin/sh' })).toBeUndefined();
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
LGTM — separator/case/.exe handling in the right order, cmd.exe correctly maps to undefined (only bash/zsh/fish completions exist), 4 new Windows-path test cases. Note for the record: the earlier Windows CI failure was a vitest worker-RPC timeout flake (2001/2001 tests passed), green on rerun.

PR #274, opened 2026-07-22
fix(browser): catch malformed URL TypeError and map to exit 5 VALIDATION_ERROR

### LINKED ISSUE #276: fix(browser): catch malformed URL TypeError and map to exit 5 VALIDATION_ERROR
### Problem
In `src/lib/browser.ts`, `openInBrowser(url)` calls `new URL(url)` without a `try/catch` block. When passed a malformed URL string (e.g., `"not a url"`), Node.js throws an un-guarded `TypeError: Invalid URL`, crashing the process with exit code 1 instead of returning a typed `VALIDATION_ERROR` (exit code 5).

### Proposed Fix
Wrap `new URL(url)` in a `try/catch` block and throw `localValidationError('url', 'must be a valid http(s) URL', undefined, 'field')`.

sandman-sh:
Closes #276

## Summary

Fixes un-guarded `new URL(url)` call in `openInBrowser` ([src/lib/browser.ts](file:///e:/project/testsprite%20cli/src/lib/browser.ts)) so malformed URLs return a clean CLI `VALIDATION_ERROR` (exit code 5) instead of crashing with an unhandled Node.js `TypeError`.

### Problem
When `openInBrowser(url)` receives a malformed or unparseable URL string (e.g. `"not a url"`), `new URL(url)` throws a raw Node.js `TypeError: Invalid URL`.
Because this call was not wrapped in a `try/catch` block:
- The process crashed with an unhandled exception (**exit code 1**) and printed a raw V8 stack trace.
- Non-HTTP schemes like `file://` already threw `VALIDATION_ERROR` (exit 5), creating an inconsistent error contract for bad inputs.

### Solution
Wrap `new URL(url)` in a `try/catch` block in `openInBrowser`:
```typescript
try {
  parsed = new URL(url);
} catch {
  throw localValidationError('url', 'must be a valid http(s) URL', undefined, 'field');
}
```

### Test Coverage
Updated [src/lib/browser.test.ts](file:///e:/project/testsprite%20cli/src/lib/browser.test.ts) to assert that malformed URL inputs throw `ApiError` with `exitCode = 5`. All unit tests pass.

### DIFF
--- a/src/lib/browser.test.ts
+++ b/src/lib/browser.test.ts
@@ -54,9 +54,16 @@ describe('openInBrowser', () => {
     expect(exec).not.toHaveBeenCalled();
   });
 
-  it('throws on a malformed URL', () => {
+  it('refuses a malformed URL with exit 5 before spawning', () => {
     const exec = vi.fn();
-    expect(() => openInBrowser('not a url', { exec })).toThrow();
+    let error: unknown;
+    try {
+      openInBrowser('not a url', { exec });
+    } catch (err) {
+      error = err;
+    }
+    expect(error).toBeInstanceOf(ApiError);
+    expect((error as ApiError).exitCode).toBe(5);
     expect(exec).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
LGTM — try-block scoped to the single `new URL()` call, error shape matches the adjacent protocol branch (VALIDATION_ERROR → exit 5), test asserts the exact exit code and the no-spawn guard. Correct and minimal.

Issue #83, opened 2026-07-01
[Hackathon] Bound Response Buffering (Large Histories/Steps Load Fully into Heap)


crypticsaiyan:
### Discord Username / User ID

uisupersaiyan3

### Summary
 
Every non-streaming HTTP response is fully materialized in memory via
`await response.json()`. For most requests this is fine, but `test result
--history` pagination and `getRun({ includeSteps: true })` can return large
payloads that are read entirely into memory before any processing or
rendering occurs.
 
---
 
### Root Cause
 
The HTTP client's typed read path treats every response uniformly via
`response.json()`, which requires the full body to be buffered and parsed
before any part of it is available to the caller. This is appropriate for
small, bounded responses, but history and step data have no inherent upper
bound on size, so the same buffering strategy applies regardless of how
large the underlying payload is.
 
---
 
### Affected Code
 
| File | Lines | Behavior |
|---|---|---|
| `src/lib/http.ts` | 497 | `response.json()` used for all typed reads, with no size bound |
| `src/lib/http.ts` | 368-383 | `getRun` step inclusion path, subject to the same unbounded read |
| `src/lib/pagination.ts` | (history pagination) | Used by `test result --history`, accumulates pages without a bound |
 
For comparison, `test code get` already streams its payload via `bundle.ts`.
The JSON read path used by history and step data does not have an
equivalent streaming or bounding mechanism.
 
---
 
### Impact
 
A run with a long step array, or a `test result --history` invocation over a
deep history window, causes heap usage to grow in proportion to the size of
the response, with no upper bound. In the worst case, this can mean loading
an entire large history or step set into memory before printing a single
row, which is both slower to first output and more memory-intensive than
necessary.
 
This is a robustness concern rather than a confirmed incident: the failure
mode is degraded performance and elevated memory use under large N, not a
crash under typical usage.

### Proposed Resolution
 
Choose one of the following approaches for the history and step read paths:
 
- **Bounded reads:** Introduce a documented ceiling on the number of
  accumulated history rows or step entries, combined with a `--limit` flag
  so callers can explicitly request a smaller window.
- **Incremental rendering:** Stream and render history pages as they arrive
  rather than buffering the full result set before printing, so memory use
  stays proportional to a single page rather than the full history.
Either approach should ensure memory use no longer scales unbounded with
response size for these two read paths.

### Confirmations

- [x] I have searched existing issues and this is not a duplicate.
- [x] I have provided my Discord identity above for reward coordination.

[...]

PR #281, opened 2026-07-24
fix(http): bound the size of buffered JSON responses

Yazan-O:
Addresses #83 — this fixes the buffered-read choke point (rows 1 and 2 of that issue's affected-code table). The page-accumulation item in `pagination.ts` is tracked separately on #83, so this PR deliberately does not close it.

## What

`HttpClient.request()` reads every successful response with `await response.json()`, which buffers the **entire** body into memory before parsing, with no upper bound. A large `test result --history` page, or a run with a long `steps[]` array (`getRun` with `includeSteps`), grows the heap in proportion to the payload.

This adds a bounded read at that single choke point:

- A declared `Content-Length` over the cap is rejected before the body is read.
- Chunked bodies (no `Content-Length`) are bounded by counting bytes as they stream, stopping once the cap is crossed.
- On exceed, it throws a typed `PAYLOAD_TOO_LARGE` error (exit 5 — the same code the backend already returns for oversized *request* bodies), with a `nextAction` pointing at `--page-size` / `--since`.
- Decoding happens once, at the end, so multi-byte UTF-8 sequences that straddle a chunk boundary still decode correctly.

The cap is a new `HttpClientOptions.maxResponseBytes` (default **64 MiB** — far above any legitimate metadata/history/steps response), injectable for tests and future env/flag wiring.

## Scope

Focused on the unbounded typed-JSON read, which is the root cause described in #83. Pagination page-count bounding is handled separately (#248), so this does **not** touch `pagination.ts`. Error-response bodies (small) keep using the existing `safeReadJson` path. Happy to widen the scope if you'd prefer.

The default cap value and whether to make it env-configurable are easy to adjust to your preference.

## Tests (`src/lib/http.test.ts`)

- Rejects an over-cap response via `Content-Length`.
- Rejects an over-cap chunked response that has no `Content-Length`.
- Reads a within-cap response unchanged.
- Updates the existing "timeout during body read" test to stall at the new read site (a body stream that errors) — same intent, still classified as `RequestTimeoutError`.

`npm run typecheck`, `npm run lint`, `npm run format:check`, and `src/lib/http.test.ts` all pass locally.

## Summary by CodeRabbit

* **New Features**
  * Added a response-size guard for successful JSON bodies (default limit: 64 MiB, configurable via `maxResponseBytes`).
  * Requests that exceed the cap are now rejected with a dedicated payload-too-large (`PAYLOAD_TOO_LARGE`) error including request details.

* **Bug Fixes**
  * Improved detection of oversized streamed/chunked responses while downloading (not just by declared size).
  * Correctly classifies timeouts that occur during response body reading.
  * Preserves existing behavior for interrupted requests and malformed successful responses.

### DIFF
--- a/src/lib/http.test.ts
+++ b/src/lib/http.test.ts
@@ -33,6 +33,7 @@ function makeClient(
     apiKey?: string | null;
     onDebug?: (e: DebugEvent) => void;
     onServerVersion?: (info: { minVersion?: string }) => void;
+    maxResponseBytes?: number;
   } = {},
 ): HttpClient {
   const apiKey = 'apiKey' in options ? (options.apiKey ?? undefined) : 'sk-test';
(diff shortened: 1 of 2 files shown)


zeshi-du:
@Yazan-O — heads-up on the Season 3 CLI Improvement Bonus coordination step: please join the TestSprite Discord and reach out to **Kairui | TestSprite** by **2026-07-29 23:59 UTC**. Slots not claimed by then may be reallocated.

https://discord.com/invite/GXWFjCe4an

zeshi-du (approved):
Approving and merging. The bounded read at the single `HttpClient.request()` choke point is the right place for this — `Content-Length` rejected before the body is touched, and chunked bodies counted as they stream so an unbounded response can't grow the heap past the cap either. That covers both rows 1 and 2 of #83's affected-code table (`response.json()` and the `getRun`/`includeSteps` path).

Two notes:

1. The `Unit Tests (Windows)` failure was a flake in a large reporter-heavy spec, not your change — I re-ran it and it passes. Sorry that a red leg made this look like your problem for three weeks.
2. I've replaced `Closes #83` with `Addresses #83` in the description. #83's third row is `pagination.ts` accumulating pages, and that's a different bound: `MAX_AUTO_PAGES` caps the page *count*, not the accumulated in-memory size, so auto-pagination over a long history can still grow the heap even with this merged. Keeping #83 open for that piece rather than letting the merge silently close it half-done.

Thanks — this is a tidy fix in exactly one place.

PR #289, opened 2026-07-26
fix(test): enforce diff dry-run exit code

### LINKED ISSUE #286: Dry-run emits a verdict-changing diff but returns before the documented exit-1 gate
## Summary

`test diff --dry-run` emits a sample whose verdict changes from passed to failed but exits successfully. The same logical result on the real path exits 1, contradicting both the command help and official exit contract.

Validation source commit: `60d55e47883e924ff1a5f0ab1fe67a25bc3b0877`.

## Code path

- `DOCUMENTATION.md:229-236`: Documentation specifies exit 1 for differing verdicts and provides a dry-run invocation.
- `src/commands/test.ts:3768-3796`: Dry-run constructs passed versus failed with verdictChanged true, then returns before the exit gate.
- `src/commands/test.ts:3861-3870`: The real path enforces exit 1 only after the dry-run early return.
- `src/commands/test.ts:8215-8222`: Command help repeats the exit contract and directly awaits runDiff.
- `src/commands/test.test.ts:2979-3054`: Tests require exit 1 for a real verdict flip but require the dry-run flip to resolve.
- Repository-local validation artifact: Minimal bounded differential witness.

## Steps to reproduce

This report is based on a source-control-flow validation against the commit above.

1. Check out the source commit listed in this report.
2. Inspect the code path and contract anchors listed above.
3. Exercise the described boundary/state path: Dry-run emits a verdict-changing diff but returns before the documented exit-1 gate.
4. Compare the observed behavior with the expected contract below.

Validation note: From the snapshot root, execute `<local validation artifact>`. It verifies the official 0/1 contract, the dry-run passed-to-failed sample with verdictChanged=true, its early successful return, the later real-path CLIError(1), and the differential test expectations. The bounded control flow proves dry-run resolves and therefore exits 0 for a reported verdict difference.

## Expected behavior

The implementation should preserve the repository contract described by the documentation, public type, parser, serializer, or sibling implementation anchors above.

## Actual behavior

`test diff --dry-run` emits a sample whose verdict changes from passed to failed but exits successfully. The same logical result on the real path exits 1, contradicting both the command help and official exit contract.

## Existing coverage

I did not find an existing issue or PR title that appears to cover this exact root cause in the pre-submission check.

## Suggested fix

Test diff dry-run exit-status gate

## Suggested tests

- Add a regression test for: Dry-run emits a verdict-changing diff but returns before the documented exit-1 gate.
- Include the boundary value or state transition described above so the old behavior fails before the fix.

---
Submitted with Codex.

zengxm1979:
﻿## Summary

Fixes the `test diff --dry-run` path so it preserves the documented CI-gate behavior when the emitted sample shows a verdict change.

## Root cause

The real `test diff` path prints the diff and throws `CLIError(..., 1)` when `verdictChanged` is true. The dry-run path prints a canned sample with `verdictChanged: true`, but returned successfully before reaching the same exit-1 gate.

## Change

- After printing the dry-run sample, throw the same verdict-difference `CLIError` when `sample.verdictChanged` is true.
- Update the dry-run unit test to assert both behaviors: JSON output is still emitted offline, and the command rejects with `exitCode: 1`.

## Workflow note

Issue #286 has been claimed with `/assign` per the contribution workflow.

## Validation

- `npm test -- src/commands/test.test.ts -t "runDiff"`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`

All checks passed locally on Node v24.15.0 / npm 11.12.1.

Closes #286

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -3295,10 +3295,12 @@ describe('runDiff', () => {
     expect(errs.join('\n')).toContain('different tests');
   });
 
-  it('--dry-run returns the canned sample fully offline (no credentials, no fetch)', async () => {
-    // Dry-run must not require credentials or hit the network — it returns a
-    // canned CliRunDiff so `--dry-run` shows the shape offline.
-    const diff = await runDiff(
+  it('--dry-run prints the canned sample fully offline and exits 1 when verdicts differ', async () => {
+    // Dry-run must not require credentials or hit the network. It still emits a
+    // canned CliRunDiff so `--dry-run` shows the shape offline, then follows the
+    // same verdict-change exit contract as the real path.
+    const out: string[] = [];
+    const rejection = await runDiff(
       {
         profile: 'default',
         output: 'json',
@@ -3307,8 +3309,15 @@ describe('runDiff', () => {
         runA: 'run_aaa',
         runB: 'run_bbb',
       },
-      { stdout: () => undefined, stderr: () => undefined },
-    );
+      { stdout: line => out.push(line), stderr: () => undefined },
+    ).catch((error: unknown) => error);
+    expect(rejection).toMatchObject({ exitCode: 1 });
+    const diff = JSON.parse(out.join('')) as {
+      runA: { runId: string };
+      runB: { runId: string };
+      verdictChanged: boolean;
+      changedSteps: Array<{ stepIndex: number; statusA: string; statusB: string }>;
+    };
     expect(diff.runA.runId).toBe('run_aaa');
     expect(diff.runB.runId).toBe('run_bbb');
     expect(diff.verdictChanged).toBe(true);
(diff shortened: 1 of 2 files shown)


zeshi-du:
Closing this as superseded, and I'm sorry — you wrote a correct fix for a bug that had already been fixed, and you had no way to know.

`test diff`'s exit contract was fixed internally on 2026-07-17 via a shared `enforceDiffExitContract()` helper called from both the dry-run branch and the real path. That predates issue #286 by 8 days; the issue was filed against the v0.3.0 release snapshot (`60d55e47`), which is what the public repo carried at the time.

That's a publishing gap on our side, not a mistake on yours. We're adding a "which commit did you verify against" field to the issue templates so triage catches this before anyone writes code.

Your other PR, #290, fixes a defect that is genuinely still open — I've approved its CI and it's next in the merge queue.

PR #301, opened 2026-08-03
fix(credentials): restore Windows ACL inheritance to fix EPERM on read

### LINKED ISSUE #304: bug: fix(credentials): Windows credentials file unreadable after setup — EPERM on read
### What happened?

After running `testsprite setup`, all subsequent commands fail with:

    Error: EPERM: operation not permitted, open '~/.testsprite/credentials'

The file is created successfully by setup, but becomes immediately unreadable.

### Steps to reproduce

```shell
1. testsprite setup --from-env --yes
2. testsprite doctor
```

### Relevant output

```shell
Error: EPERM: operation not permitted, open 'C:\Users\<username>\.testsprite\credentials'
```

### CLI version

0.4.0

### Node.js version

v24.14.0

### Operating system

Windows 11 Pro

### Confirmations

- [x] I am on the latest published version, or have noted why I cannot upgrade.
- [x] I have removed any secrets (API keys, credentials, private URLs) from this report.

aythin:
## What does this PR do?

`ensureWindowsRestrictiveAcl` previously called:

```
icacls credentials /inheritance:r /grant:r USERNAME:F
```

`/inheritance:r` strips **all** inherited ACEs and leaves the file protected solely by the `USERNAME` env-var grant. On Windows with Microsoft Accounts or domain-joined machines, `USERNAME` may not resolve to the same SID that owns the file — the resulting ACL locks out the file owner, causing every subsequent credential read to fail with `EPERM: operation not permitted`.

**Fix:** call `icacls /reset` first (re-enables inheritance from the parent directory), then `/grant:r USERNAME:F` as an explicit belt-and-suspenders Full Control entry.

Verified on Windows 11 Pro:
- `testsprite setup --from-env --yes` + `testsprite doctor` previously failed immediately with EPERM
- After this fix, `testsprite doctor` reports all checks passed ✅

## Related issue

None — small bug fix, no issue required per CONTRIBUTING.md.

## Type of change

## Checklist

- [x] Commits follow Conventional Commits (`fix(credentials): ...`).
- [x] `npm test` passes and coverage stays at or above the 80% gate.
- [x] New behavior is covered by unit tests (updated `credentials.test.ts` to expect two `icacls` calls).
- [x] User-facing changes are reflected in `README.md` / `DOCUMENTATION.md` where relevant. (N/A — internal Windows ACL fix)

## Notes for reviewers

Only two files changed: `src/lib/credentials.ts` (the fix) and `src/lib/credentials.test.ts` (updated test assertions to match the new two-call sequence: `/reset` then `/grant:r`).

## Bug Fixes
- Improved Windows credential file permissions by resetting inherited access before applying restrictive access controls.
- Strengthened protection for stored credentials by granting access through Windows OWNER RIGHTS rather than relying on the logged-in username.
- Added clearer warning handling when permission updates fail or return unsuccessful results.

## Tests
- Expanded Windows permission checks to cover reset and access-grant failures.
- Verified the expected command sequence and secure execution behavior.

Closes #304

### DIFF
--- a/src/lib/credentials.test.ts
+++ b/src/lib/credentials.test.ts
@@ -219,27 +219,17 @@ describe('ensureRestrictiveMode', () => {
     // First call: /reset re-enables inheritance from the parent directory so the
     // owner can always access the file (fixes EPERM on Microsoft Account / domain
     // account machines where USERNAME does not resolve to the file-owner SID).
-    expect(spawn).toHaveBeenNthCalledWith(
-      1,
-      'icacls',
-      [credentialsPath, '/reset'],
-      {
-        shell: false,
-        stdio: 'ignore',
-        windowsHide: true,
-      },
-    );
+    expect(spawn).toHaveBeenNthCalledWith(1, 'icacls', [credentialsPath, '/reset'], {
+      shell: false,
+      stdio: 'ignore',
+      windowsHide: true,
+    });
     // Second call: /grant:r adds an explicit Full Control entry as belt-and-suspenders.
-    expect(spawn).toHaveBeenNthCalledWith(
-      2,
-      'icacls',
-      [credentialsPath, '/grant:r', 'alice:F'],
-      {
-        shell: false,
-        stdio: 'ignore',
-        windowsHide: true,
-      },
-    );
+    expect(spawn).toHaveBeenNthCalledWith(2, 'icacls', [credentialsPath, '/grant:r', 'alice:F'], {
+      shell: false,
+      stdio: 'ignore',
+      windowsHide: true,
+    });
     expect(spawn).toHaveBeenCalledTimes(2);
   });
 
--- a/src/lib/credentials.test.ts
+++ b/src/lib/credentials.test.ts
@@ -205,49 +205,121 @@ describe('ensureRestrictiveMode', () => {
     expect(() => ensureRestrictiveMode(credentialsPath)).not.toThrow();
   });
 
-  it('tightens the Windows ACL with icacls instead of POSIX chmod', () => {
+  it('tightens the Windows ACL with icacls using /reset then /inheritance:r /grant:r *S-1-3-4:F', () => {
     mkdirSync(tmpRoot, { recursive: true });
     writeFileSync(credentialsPath, 'data', { mode: 0o666 });
     const spawn = vi.fn(() => ({ status: 0, signal: null, output: [], pid: 123 })) as never;
 
     ensureRestrictiveMode(credentialsPath, {
       platform: 'win32',
-      env: { USERNAME: 'alice' } as NodeJS.ProcessEnv,
       spawnSync: spawn,
     });
 
-    // First call: /reset re-enables inheritance from the parent directory so the
-    // owner can always access the file (fixes EPERM on Microsoft Account / domain
-    // account machines where USERNAME does not resolve to the file-owner SID).
+    // Step 1: /reset clears broken inheritance state.
     expect(spawn).toHaveBeenNthCalledWith(1, 'icacls', [credentialsPath, '/reset'], {
       shell: false,
       stdio: 'ignore',
       windowsHide: true,
     });
-    // Second call: /grant:r adds an explicit Full Control entry as belt-and-suspenders.
-    expect(spawn).toHaveBeenNthCalledWith(2, 'icacls', [credentialsPath, '/grant:r', 'alice:F'], {
-      shell: false,
-      stdio: 'ignore',
-      windowsHide: true,
-    });
+    // Step 2: /inheritance:r removes inherited ACEs and /grant:r grants Full Control
+    // to the OWNER RIGHTS SID (*S-1-3-4), avoiding dependency on USERNAME resolution.
+    expect(spawn).toHaveBeenNthCalledWith(
+      2,
+      'icacls',
+      [credentialsPath, '/inheritance:r', '/grant:r', '*S-1-3-4:F'],
+      {
+        shell: false,
+        stdio: 'ignore',
+        windowsHide: true,
+      },
+    );
     expect(spawn).toHaveBeenCalledTimes(2);
   });
 
-  it('warns on Windows when credentials ACL tightening cannot run', () => {
+  it('warns on Windows when icacls /reset fails with an error', () => {
     mkdirSync(tmpRoot, { recursive: true });
     writeFileSync(credentialsPath, 'data');
     const warnings: string[] = [];
-    const spawn = vi.fn(() => ({ status: 0, signal: null, output: [], pid: 123 })) as never;
+    const spawn = vi.fn(() => ({
+      error: new Error('spawnSync icacls ENOENT'),
+      status: null,
+      signal: null,
+      output: [],
+      pid: 123,
+    })) as never;
 
     ensureRestrictiveMode(credentialsPath, {
       platform: 'win32',
-      env: {} as NodeJS.ProcessEnv,
       spawnSync: spawn,
       warn: line => warnings.push(line),
     });
 
-    expect(spawn).not.toHaveBeenCalled();
-    expect(warnings.join('\n')).toContain('credentials file permissions were not tightened');
+    expect(spawn).toHaveBeenCalledTimes(1);
+    expect(warnings.join('\n')).toContain(
+      'icacls failed while resetting credentials file permissions',
+    );
+  });
+
+  it('warns on Windows when icacls /reset exits non-zero', () => {
+    mkdirSync(tmpRoot, { recursive: true });
+    writeFileSync(credentialsPath, 'data');
+    const warnings: string[] = [];
+    const spawn = vi.fn(() => ({ status: 1, signal: null, output: [], pid: 123 })) as never;
+
+    ensureRestrictiveMode(credentialsPath, {
+      platform: 'win32',
+      spawnSync: spawn,
+      warn: line => warnings.push(line),
+    });
+
+    expect(spawn).toHaveBeenCalledTimes(1);
+    expect(warnings.join('\n')).toContain('icacls /reset exited with status 1');
+  });
+
+  it('warns on Windows when /reset succeeds but /grant:r fails with an error', () => {
+    mkdirSync(tmpRoot, { recursive: true });
+    writeFileSync(credentialsPath, 'data');
+    const warnings: string[] = [];
+    const spawn = vi
+      .fn()
+      .mockReturnValueOnce({ status: 0, signal: null, output: [], pid: 123 })
+      .mockReturnValueOnce({
+        error: new Error('permission denied'),
+        status: null,
+        signal: null,
+        output: [],
+        pid: 123,
+      }) as never;
+
+    ensureRestrictiveMode(credentialsPath, {
+      platform: 'win32',
+      spawnSync: spawn,
+      warn: line => warnings.push(line),
+    });
+
+    expect(spawn).toHaveBeenCalledTimes(2);
+    expect(warnings.join('\n')).toContain(
+      'icacls failed while tightening credentials file permissions',
+    );
+  });
+
+  it('warns on Windows when /reset succeeds but /grant:r exits non-zero', () => {
+    mkdirSync(tmpRoot, { recursive: true });
+    writeFileSync(credentialsPath, 'data');
+    const warnings: string[] = [];
+    const spawn = vi
+      .fn()
+      .mockReturnValueOnce({ status: 0, signal: null, output: [], pid: 123 })
+      .mockReturnValueOnce({ status: 5, signal: null, output: [], pid: 123 }) as never;
+
+    ensureRestrictiveMode(credentialsPath, {
+      platform: 'win32',
+      spawnSync: spawn,
+      warn: line => warnings.push(line),
+    });
+
+    expect(spawn).toHaveBeenCalledTimes(2);
+    expect(warnings.join('\n')).toContain('icacls exited with status 5');
   });
 
   // POSIX-only premise: Windows has no 0644/0600 distinction to downgrade.
--- a/src/lib/credentials.test.ts
+++ b/src/lib/credentials.test.ts
@@ -216,15 +216,31 @@ describe('ensureRestrictiveMode', () => {
       spawnSync: spawn,
     });
 
-    expect(spawn).toHaveBeenCalledWith(
+    // First call: /reset re-enables inheritance from the parent directory so the
+    // owner can always access the file (fixes EPERM on Microsoft Account / domain
+    // account machines where USERNAME does not resolve to the file-owner SID).
+    expect(spawn).toHaveBeenNthCalledWith(
+      1,
       'icacls',
-      [credentialsPath, '/inheritance:r', '/grant:r', 'alice:F'],
+      [credentialsPath, '/reset'],
       {
         shell: false,
         stdio: 'ignore',
         windowsHide: true,
       },
     );
+    // Second call: /grant:r adds an explicit Full Control entry as belt-and-suspenders.
+    expect(spawn).toHaveBeenNthCalledWith(
+      2,
+      'icacls',
+      [credentialsPath, '/grant:r', 'alice:F'],
+      {
+        shell: false,
+        stdio: 'ignore',
+        windowsHide: true,
+      },
+    );
+    expect(spawn).toHaveBeenCalledTimes(2);
   });
 
   it('warns on Windows when credentials ACL tightening cannot run', () => {
--- a/src/lib/credentials.ts
+++ b/src/lib/credentials.ts
@@ -77,7 +77,6 @@ interface CredentialsLock {
 
 interface RestrictiveModeOptions {
   platform?: NodeJS.Platform;
-  env?: NodeJS.ProcessEnv;
   spawnSync?: (
     command: string,
     args: readonly string[],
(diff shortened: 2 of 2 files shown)


zeshi-du:
Thanks for this — the diagnosis is exactly right, and it's a regression we introduced: an earlier Windows ACL hardening change runs `icacls <path> /inheritance:r /grant:r %USERNAME%:F` with no `/reset`, so on Microsoft-Account or domain-joined Windows the explicit `USERNAME` grant may not resolve to the SID that owns the file and the owner gets locked out of their own credentials. `/reset` before the grant is the right shape.

Your CI had never been approved to run (first-time-contributor gate — our fault, not yours). I've approved it, and now that it has run there's one thing to fix:

- **`Lint & Format` fails.** Run `npm run lint:fix && npm run format` and commit the result.

Everything else is green, including the Windows leg, which is the one that matters here.

One request while you're in there: please make sure there's a unit test that pins the `icacls` argument order — the existing `spawnSync` mock in `credentials.test.ts` gives you the seam. This code path can only fail on real Windows with a non-local account, so the test is the only thing that will stop the next person from re-introducing `/inheritance:r` without `/reset`.

Push those and I'll merge. This is a customer-visible break for Windows users, so it's going out in the next release.

zeshi-du:
Correcting my own steer from 8/13: "`/reset` before the grant is the right shape" was true as far as it went, but I said "push those and I'll merge" without checking what `/reset` + `/grant:r` actually leaves behind — and it doesn't leave the file restrictive. That gap is mine, not yours. Here's what actually needs to change before I can merge.

**What `/reset` + `/grant:r` alone does, and why it isn't equivalent to the `/inheritance:r` it replaces:**

- `icacls <path> /reset` (`credentials.ts:242`) clears the "protected" flag and re-enables inheritance from the parent directory. It adds no ACE of its own.
- `icacls <path> /grant:r <user>:F` (`credentials.ts:248`) replaces only `<user>`'s explicit ACE. It doesn't strip the inherited ACEs `/reset` just turned back on, and it doesn't restore the protected flag.
- Net effect: the file is no longer decoupled from its parent directory's ACL — it inherits whatever the parent grants, and will keep tracking any future change to that parent ACL. That's the opposite of what this function's own docstring promises ("Windows hosts use ACL tightening via icacls", `credentials.ts:207`), and it's now a real asymmetry with the POSIX branch (`credentials.ts:216`), which still does `chmod(0600)`.
- On a standard single-user Windows profile this is a modest gap — inherited ACEs there are typically just owner + SYSTEM + Administrators. On a domain-joined or GPO-managed machine it's wider, since inherited ACEs routinely include broader groups there. This is a hardening regression, not a live exploit — but it's exactly the thing `ensureWindowsRestrictiveAcl` exists to prevent, so I can't take it as-is.

**Root cause — not walking this back:** your diagnosis is correct. `%USERNAME%` doesn't reliably resolve to the SID that owns the file on Microsoft-Account or domain-joined machines, so the old `/inheritance:r /grant:r %USERNAME%:F` could strip every inherited ACE and then grant to a SID nobody has, leaving an ACL nobody can read — exactly the EPERM in #304. `/reset` first is the right move for unwinding that broken state. It just can't be the last step.

**What I need:** keep `/reset` as the cleanup, then re-tighten with a grant target that's guaranteed to resolve. Pick whichever of these — I'll take a working version of any of them:

**(a)** Grant to the well-known `OWNER RIGHTS` SID instead of a username. It always resolves and always means "whoever owns this file":
```
icacls <path> /reset
icacls <path> /inheritance:r /grant:r *S-1-3-4:F
```
This is closest to what the docstring already claims, and it removes the dependency on `USERNAME` resolving to anything — which is the actual root cause. My preference, but not a requirement.

**(b)** Resolve the real owning SID first (e.g. `[System.Security.Principal.WindowsIdentity]::GetCurrent().User.Value` or `whoami /user`) and grant to that SID literal instead of the `USERNAME` string.

**(c)** Grant first, then `/inheritance:r`, then read the resulting ACL back and confirm the intended principal has `F` before returning — rolling back to the inherited state (re-`/reset`) if the tightening left the file unreadable, instead of leaving it silently broader.

**Two more things while you're in there:**

1. The first `run('icacls', [path, '/reset'], ...)` call (`credentials.ts:242`) discards its return value — only the second call's result is checked. CodeRabbit flagged exactly this on this PR on 2026-08-03 and it's still open on the current head: if `/reset` fails, execution falls straight through to the grant with no warning. Check `result.error` / `result.status` on both calls the same way.
2. There's no test for `/reset` failing, or `/reset` succeeding while `/grant:r` fails. The existing "warns on Windows when credentials ACL tightening cannot run" test only covers a missing `USERNAME` env var, where `spawnSync` is never called at all — it doesn't exercise the icacls-failure path either call could actually hit.

**Separate, factual, not something to fix in this PR:** `mutateCredentialsFile`'s `mkdirSync(dirname(path), { recursive: true, mode: 0o700 })` (`credentials.ts:283`) — the `mode` option is a documented no-op on Windows, so the parent directory was never actually tightened by us on this platform either. Just context for why the file-level ACL is the only real control point here; not blocking.

**On CI:** your 2026-08-18 push was a pure Prettier reflow, no logic change. Full CI had never actually run on that head — the fork-approval gate re-arms on every push from a first-time contributor, so your checks tab was empty through no fault of yours. I've approved the run now. Two failures in it are ours, not yours: `Dependency Review` (org-level dependency graph is off org-wide, fails on every PR) and `ESLint Security (changed files)` (pre-existing lint backlog; the fix ships in our source repo on the next release). `Unit Tests (Windows)` also failed on the first attempt, but on an unrelated timeout in `test.test.ts` (`runFailureGet > in-place rewrite removes stale top-level files` — `Test timed out in 5000ms`, in a filesystem-heavy test whose sibling took 2562ms on the same runner). Your `credentials.test.ts` passed clean in that same run (25 passed, 1 skipped, 0 failed), and this PR touches nothing that test reaches. I re-ran the Windows job to be sure rather than asserting it — **it passed**, so that was a runner-speed flake, not your change. Everything except the two infrastructure checks above is now green.

Push the ACL fix plus the two follow-ups above and I'll take another look.

aythin:
Hi @zeshi-du, thank you for the exceptionally thorough review — and for owning the gap in the earlier steer. No apology needed at all; your analysis of what `/reset` + `/grant:r` leaves behind made the fix straightforward to land.

I went with **option (a)**, since it addresses the root cause directly:

- `icacls <path> /reset` — unwinds any broken inheritance state from the previous `/inheritance:r` attempt (the EPERM from #304)
- `icacls <path> /inheritance:r /grant:r *S-1-3-4:F` — decouples the file from the parent ACL and grants Full Control to the well-known OWNER RIGHTS SID, removing the dependency on `%USERNAME%` resolving to the owner's SID entirely

Both follow-ups are also addressed:

1. Both `icacls` calls now check `result.error` and `result.status`, with distinct warning messages for the `/reset` and grant phases (the first call previously discarded its result — thanks to CodeRabbit for flagging that originally).
2. `credentials.test.ts` now pins the exact argument order of both calls, and adds coverage for all four failure paths: `/reset` erroring, `/reset` exiting non-zero, and the grant failing (error / non-zero) after a successful `/reset`.

I also removed the now-unused `env` field from `RestrictiveModeOptions`, since nothing reads `%USERNAME%` anymore.

Noted on `mkdirSync`'s `mode` being a no-op on Windows — thanks for the context; happy to look at that separately if it'd be useful.

zeshi-du:
Apologies for the silence — you pushed the OWNER RIGHTS SID fix on 2026-08-28 and it sat here for over two weeks. That was the exact thing we asked for on 08-19, and it's the right approach: `process.env.USERNAME` can't be relied on to resolve to the file owner's SID on a Microsoft-account or domain-joined machine, which is what turned the ACL hardening into a lockout in the first place.

Status and what's next:

- It still merges cleanly against current `main` (v0.11.0) — I re-checked today, no rebase needed.
- **Your CI has never run.** Fork PRs on this repo sit at `action_required` until a maintainer releases them, and nobody did. That's on us, not on you; it's being unblocked now, and I'll review against a green run.

Two things I expect to raise in that review, flagged early so you can decide whether to fold them in:

1. **Already-affected users aren't recovered.** Someone who ran an affected version still has a credentials file carrying an ACL they can't read. Worth either a repair path or, at minimum, a `doctor` message that says to delete `~/.testsprite/credentials` and re-run `setup`.
2. **The tightening step fails open.** If the ACL change fails after the file is created, we continue silently with default inheritance. A stderr warning would make that visible.

Neither is large. Thanks for staying with this one.

zeshi-du:
CI has finally run on this — that was our fault, the workflow was sitting unapproved for 18 days. Results: **everything passes except one real item and one that isn't yours.**

**Real, and it's a two-minute fix.** `ESLint Security (changed files)` flags six `security/detect-non-literal-fs-filename` findings in `src/lib/credentials.test.ts`:

```
313:5  mkdirSync      329:5  mkdirSync      356:5  mkdirSync
314:5  writeFileSync  330:5  writeFileSync  357:5  writeFileSync
```

These are test fixtures writing into the suite's own `mkdtempSync` directory, which is fine — the rule just can't see that. This repo's established convention is an inline disable **with a reason**, e.g. from `src/commands/doctor.tunnel.spec.ts`:

```ts
// eslint-disable-next-line security/detect-non-literal-fs-filename -- `dir` is this suite's own mkdtempSync temp dir, never user input
mkdirSync(dir, { recursive: true });
```

Add one of those above each of the six lines and the job goes green. Please don't silence the rule at config level.

**Not yours:** `Dependency Review` fails on every PR in this repo because our organisation has the dependency graph disabled. A fix is in flight internally. Ignore it.

Everything else is green, including `Unit Tests (Windows)` — which is the platform that matters most for this change.

With that and the two review points from my earlier comment (a repair path for already-bricked files, and not failing open silently when the ACL tightening fails), this merges.

aythin:
Hi @zeshi-du, thanks for unblocking CI and for flagging the review points early — all three items are addressed in the latest push:

**1. ESLint Security findings (the six `security/detect-non-literal-fs-filename` lines)**

Added the inline disable with a reason above each of the six calls in `src/lib/credentials.test.ts`, following the convention from `doctor.tunnel.spec.ts` — the reason notes the paths are the suite's own `mkdtempSync` temp dir, never user input. The rule itself is untouched at config level.

**2. Recovery for already-affected users**

Went with the doctor-message option. `testsprite doctor` now detects the bricked state directly: when the credentials file exists but the OS refuses the read (`EPERM`/`EACCES` — the signature of the locked-out ACL), it no longer dies with a raw fs error. Instead the Credentials check reports:

> credentials file exists but cannot be read (EPERM); its ACL may be locked out by an earlier CLI version — delete the file and re-run `testsprite setup`

If `TESTSPRITE_API_KEY` is set (commands still work), the check degrades to a warning instead of a failure. Covered by two new tests in `doctor.test.ts` (fail without env key, warn with it).

**3. Tightening must not fail open silently**

The current head already checks `result.error` and `result.status` on **both** `icacls` calls, and every failure path emits a stderr warning via the default warn sink. To pin that contract, `credentials.test.ts` now has a test that runs with **no injected warn callback** and asserts the `[warning] …` line lands on `process.stderr` — so a future regression to silent fail-open fails CI.

Happy to iterate on any of the above. And noted on Dependency Review — understood that it's org-side and not something I can affect from the PR.

jangjos-128:
Hello @aythin, thanks for the latest push, and apologies again on CI. 

GitHub builds `pull_request` runs from the PR's merge commit, and this PR is now conflicted, so there is no merge commit to build from and no check suite gets scheduled. So the rebase is the unblock rather than a trailing cleanup step: once the conflict clears, the runs will be created. 

The ACL work is right, and I checked the assumption it rests on rather than taking it. I built this head on Windows 11 and bricked a credentials file the way the bug does, leaving only SYSTEM on the ACE.` icacls /reset` still succeeds as the owner, so the repair path is  genuinely viable. One thing worth a comment in the code: the `F` in the OWNER RIGHTS grant is load-bearing, not incidental. After `/inheritance:r` the file carries exactly one ACE, and Full Control is what leaves `WRITE_DAC `with the owner, which is what makes a later `/reset` possible at all. If someone "tightens" that to `:R`, the file becomes unrepairable without taking ownership.

One blocker, in the doctor recovery rather than the ACL. Same machine, real unreadable file. With no env key, your feature works exactly as intended:

  ```
[FAIL] Credentials   credentials file exists but cannot be read (EPERM); ... delete the file and re-run `testsprite setup`
  [WARN] Connectivity  skipped; no API key to test with
```

  With `TESTSPRITE_API_KEY `set, which is the case the PR describes as degrading to a warning:

```
  [WARN] Credentials   ... (TESTSPRITE_API_KEY is set, so commands still work)
  [FAIL] Connectivity  GET /me failed (EPERM: operation not permitted, open '...\.testsprite\credentials')

  1 failure(s), 2 warning(s).
```

Exit code 1. `checkConnectivity `calls `makeHttpClient`, which calls `loadConfig `with the same `credentialsPath`, so the file is read a second time and throws the same EPERM. The catch folds it into the generic branch, so a filesystem error surfaces as an API failure and the report contradicts itself.

The tests miss it because `loadConfigFn `is injected into `runDoctor `only, and that seam never reaches `makeHttpClient`. Both new tests also sit outside the `writeProfile `setup the other tests use, so there's no file at `credentialsPath `for connectivity to trip over. `expect(report.failures).toBe(0)` passes for a reason that doesn't hold in production.

Smallest fix in keeping with what you've already done: let the resolved `config `reach the client, for example an optional pre-resolved config on `ClientFactoryDeps `that `makeHttpClient `uses instead of calling `loadConfig`. Then pin it with a test that drives a real unreadable file. `chmod 000` gives `EACCES `on Linux CI, which the code already accepts, so it works cross-platform.

Rebase, plus the connectivity fix and its test, and I believe this is done. Thanks for staying with it!

**Review by zeshi-du — DISMISSED** [review]
Hi @aythin — good catch on the `icacls /inheritance:r` root cause, that's a solid fix. One gap remains in the new `doctor.ts` recovery path:

- When credentials are locked (EPERM/EACCES) but `TESTSPRITE_API_KEY` is set, `runDoctor` builds a usable config and reports a Credentials warning — but `checkConnectivity` → `makeHttpClient` re-reads the same locked file via `loadConfig` instead of reusing the config `runDoctor` already resolved, so connectivity still fails and `doctor` exits 1 even with a valid env key. Please thread the already-resolved config into the connectivity check instead of re-reading.
- Please add a regression test for exactly that case (locked file + env key → doctor passes), and rebase `doctor.ts` against `main` — it currently conflicts, which is also why real CI hasn't run on this head yet.

Thanks for the detailed root-cause write-up on the ACL issue, that part is solid.

**Comment by aythin** [comment]
Hi @jangjos-128, thank you for taking the time to test this on a real Windows machine, all addressed:

- **Rebased onto current main** — the merge commit exists, so CI should schedule now. Thanks for your patience with the fork-approval friction along the way.
- **Double-read fixed** — `ClientFactoryDeps` accepts an optional pre-resolved `config`; when present the factory skips its own `loadConfig`. `runDoctor` passes its config into both `checkConnectivity` **and** `checkLocalTunnel` — while rebasing I noticed the tunnel check had the same latent double-read, so it's covered too.
- **Real-path test** — creates a credentials file, `chmod 000`s it (EACCES on Linux CI), sets `TESTSPRITE_API_KEY`, and asserts Credentials reports the warning while neither Connectivity nor Local tunnel mention the fs error; `report.failures === 0`. Skipped on Windows (chmod is a no-op there); the injected-EPERM tests stay for cross-platform detail coverage.
- **`:F` comment added** — the grant now notes that Full Control is what leaves WRITE_DAC with the owner and keeps a later `/reset` possible; `:R` would make the file unrepairable without taking ownership.

Thanks for verifying the `/reset` repair path on a real bricked file — good to know the recovery story holds where it matters.

**Comment by jangjos-128** [comment]
Hi @aythin, thank you for the quick turnaround!

The open change request predates this push — everything it asks for is in `b497b33b`, so nothing there needs action from you; we'll get the review state refreshed on our side.

I checked `b497b33b `against the code rather than the commit message and it all holds: runDoctor threads the resolved config into both `checkConnectivity `and `checkLocalTunnel`, and client-factory now does `deps.config ?? loadConfig(...)`, so the credentials file is read once. Spotting that the tunnel check had the same latent double-read was a good catch — that one wasn't on the list.

The new chmod 000 test pins the case that matters: on a genuinely unreadable file with TESTSPRITE_API_KEY set, Connectivity and Local tunnel both stay ok and report.failures === 0. It's green on Node 20 and 22, and Unit Tests (Windows) passes as well. The rebase did its job too — CI is scheduling on this head again. And thanks for the :F note in ensureWindowsRestrictiveAcl; that's the detail that keeps the file repairable.

One item left, and it's the only red check — `ESLint Security (changed files)`:

```
src/commands/doctor.test.ts:331:7  security/detect-non-literal-fs-filename 
Found chmodSync from package "node:fs" with non literal argument at index 0
```

Same convention you already applied in `credentials.test.ts`:

```
  // eslint-disable-next-line security/detect-non-literal-fs-filename -- `credentialsPath` is this suite's own mkdtempSync temp dir, never user input
  chmodSync(credentialsPath, 0o000);
```

Non-blocking, while you're in there: the new test's `expect(out).not.toContain('EACCES: operation not permitted')` can never fail — Node renders EACCES as "permission denied"; "operation not permitted" is EPERM's wording. The assertions that carry the weight `(connectivity?.status === 'ok', failures === 0)` are correct, so it's a dead line rather than a hole.

Push the lint fix and this is ready to merge. Thanks for staying with it through the CI friction — the ACL fix and the recovery path are both solid work.

Issue #282, opened 2026-07-24
fix(project): credential/auto-auth file flags bypass the file-read guard (raw ENOENT, exit 1, malformed --output json)


OkeyAmy:
## Summary

Four file-reading flags added in v0.4.0 on `project credential` and `project auto-auth` read their file with a raw `readFileSync(path, 'utf8').trim()` and **no guard**. When the file is missing (or is a directory), the CLI throws an unwrapped Node error instead of the structured validation error every other file flag produces:

- **exit `1`** (generic) instead of **`5`** (validation)
- a **malformed `--output json` envelope** — `error` is a bare string, not the `{ code, message, nextAction }` object the rest of the CLI emits, so anything parsing `--output json` breaks
- it **leaks raw fs internals** (absolute path + errno)
- it also **crashes under `--dry-run`**, which is documented not to touch the filesystem

## Affected flags

| Command | Flag | Site |
|---|---|---|
| `project credential` | `--credential-file` | `src/commands/project.ts:470` |
| `project auto-auth` | `--password-file` | `src/commands/project.ts:579` |
| `project auto-auth` | `--client-secret-file` | `src/commands/project.ts:583` |
| `project auto-auth` | `--refresh-token-file` | `src/commands/project.ts:588` |

All six use the same unguarded pattern:

```ts
credential = readFileSync(opts.credentialFile, 'utf8').trim();
```

## Reproduction

The contrast is the bug — the same class of user error (bad input to a file flag) gets two different contracts:

```console
# Guarded flag (correct): structured envelope, exit 5
$ testsprite test create --project x --type backend --name n --code-file /nope.py --output json
{
  "error": {
    "code": "VALIDATION_ERROR",
    "message": "Invalid request.",
    "nextAction": "Flag `--code-file` is invalid: file does not exist: /nope.py.",
    "requestId": "local",
    "details": { "field": "code-file", ... }
  }
}
# exit 5

# Unguarded flag (bug): bare-string envelope, exit 1, leaks fs internals
$ testsprite project credential <project-id> --type "API key" --credential-file /nope --output json
{
  "error": "ENOENT: no such file or directory, open '/nope'"
}
# exit 1
```

`--dry-run` does not help — it reads the file (and crashes) before the dry-run banner:

```console
$ testsprite project credential <project-id> --type "API key" --credential-file /nope --dry-run
Error: ENOENT: no such file or directory, open '/nope'
# exit 1
```

[...]

PR #302, opened 2026-08-05
fix(project): guard --password-file reads on create/update

### LINKED ISSUE #79: [Hackathon] invalid --since and missing --password-file crash instead of returning VALIDATION_ERROR

naufalfx805-source:
Closes #79

Scoped per @zeshi-du's [2026-07-18 status correction](https://github.com/TestSprite/testsprite-cli/issues/79#issuecomment-5009029973): the `--since` half landed in #27, and this PR takes the `--password-file` half that is still live on `main`.

## The defect

`project create` and `project update` resolve `--password-file` with a bare read:

```ts
password = readFileSync(opts.passwordFile, 'utf8').trim();
```

A path typo is the expected failure mode for a hand-typed flag, but it escapes as an unhandled Node exception rather than the CLI's typed validation error.

**Before** (`origin/main` @ fe07bc9):

```console
$ testsprite project create --type backend --name Guarded \
    --password-file /nope/missing.txt --output json
{
  "error": "ENOENT: no such file or directory, open '/nope/missing.txt'"
}
$ echo $?
1
```

Three things are wrong: exit `1` instead of `5`, an `error` that is a bare string rather than the `{ code, message, nextAction }` envelope every other command emits (so `--output json` consumers break), and the absolute path plus errno leaked to stderr.

**After** (this branch):

```console
$ testsprite project create --type backend --name Guarded \
    --password-file /nope/missing.txt --output json
{
  "error": {
    "code": "VALIDATION_ERROR",
    "message": "Invalid request.",
    "nextAction": "Flag `--password-file` is invalid: file does not exist: /nope/missing.txt.",
    "requestId": "local",
    "details": {
      "field": "password-file",
      "reason": "file does not exist: /nope/missing.txt"
    }
  }
}
$ echo $?
5
```

A directory argument (`--password-file /tmp`) previously crashed with `EISDIR`; it now returns the same envelope with `not a regular file`.

## The change

New `src/lib/secret-file.ts` exporting `readSecretFileGuarded(flag, path)`, mirroring `readCodeFileGuarded` in `src/commands/test.ts` — `statSync` first, map `ENOENT` / `EACCES` / `EPERM` / `EISDIR` and the not-a-regular-file case onto `localValidationError`, then read. Both call sites (`project.ts:213`, `project.ts:329`) route through it.

Errors report the path **as typed**, not the resolved absolute path, so no directory layout leaks into output.

## Scope: deliberately not touching `project auto-auth`

`project.ts:579` (`project auto-auth --password-file`) is the third bare site, but it belongs to #282 along with `--credential-file`, `--client-secret-file`, and `--refresh-token-file` — that issue is assigned and in progress. Fixing it here would collide.

Instead the helper takes the **flag name** as its first argument, so #282 can adopt it directly:

```ts
credential = readSecretFileGuarded('credential-file', opts.credentialFile);
```

That is the "split-out password-file guard" @zeshi-du pointed at in #248, available as a reusable unit rather than tangled with that PR's pagination work.

## Two judgement calls

- **No payload cap.** `readCodeFileGuarded` enforces `MAX_INLINE_CODE_BYTES`. Secrets are small, and adding a size ceiling to a shipped flag is a behaviour change rather than part of fixing a crash. Happy to add one if you want the parity.
- **No explicit BOM strip.** `U+FEFF` is ECMAScript whitespace, so the existing `.trim()` already removes a BOM written by PowerShell 5.1's default `Set-Content -Encoding utf8`. There is a regression test pinning that, so it cannot silently regress if the trim is ever refactored.

## Dry-run

Unchanged — both commands already return before password resolution, and the existing `P7 — dry-run with --password-file does not read the filesystem` test still passes.

## Tests

15 new tests: 11 unit (`src/lib/secret-file.test.ts`) covering happy path, trimming, BOM, interior whitespace, relative-path resolution, empty file, missing file (code + exit + message + caller-supplied flag name + no absolute-path leak), and the directory case; 4 command-level (`src/commands/project.test.ts`) asserting `runCreate` / `runUpdate` reject with `VALIDATION_ERROR` exit 5 **before any network call**, that `nextAction` names the flag, and that a valid file is still read and sent.

```
npm test        → 59 files, 2043 passed | 2 skipped, 0 failing
npm run typecheck → clean
npx eslint      → clean
npx prettier --check → clean
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of password files during project creation and updates.
  * Missing, inaccessible, empty, or invalid password files now produce clear validation errors before any network request.
  * Password files are validated as regular files, with paths resolved consistently.
  * Existing valid password files continue to be read and submitted correctly.

### DIFF
--- a/src/commands/project.test.ts
+++ b/src/commands/project.test.ts
@@ -1044,6 +1044,122 @@ describe('runUpdate', () => {
   });
 });
 
+describe('#79 — an unreadable --password-file is a validation error, not a crash', () => {
+  const missing = join(tmpdir(), 'testsprite-issue-79-absent-password-file');
+
+  it('runCreate rejects a missing file with VALIDATION_ERROR (exit 5) before the network', async () => {
+    const { credentialsPath } = makeCreds();
+    const fetchImpl = vi.fn(async () => {
+      throw new Error('should not hit network');
+    });
+
+    await expect(
+      runCreate(
+        {
+          profile: 'default',
+          output: 'json',
+          debug: false,
+          type: 'backend',
+          name: 'Guarded',
+          passwordFile: missing,
+        },
+        {
+          credentialsPath,
+          fetchImpl: fetchImpl as unknown as typeof fetch,
+          stdout: () => {},
+          stderr: () => {},
+        },
+      ),
+    ).rejects.toMatchObject({ code: 'VALIDATION_ERROR', exitCode: 5 });
+
+    expect(fetchImpl).not.toHaveBeenCalled();
+  });
+
+  it('runUpdate rejects a missing file with VALIDATION_ERROR (exit 5) before the network', async () => {
+    const { credentialsPath } = makeCreds();
+    const fetchImpl = vi.fn(async () => {
+      throw new Error('should not hit network');
+    });
+
+    await expect(
+      runUpdate(
+        {
+          profile: 'default',
+          output: 'json',
+          debug: false,
+          projectId: 'proj_guarded',
+          passwordFile: missing,
+        },
+        {
+          credentialsPath,
+          fetchImpl: fetchImpl as unknown as typeof fetch,
+          stdout: () => {},
+          stderr: () => {},
+        },
+      ),
+    ).rejects.toMatchObject({ code: 'VALIDATION_ERROR', exitCode: 5 });
+
+    expect(fetchImpl).not.toHaveBeenCalled();
+  });
+
+  it('names the flag in nextAction instead of leaking a raw ENOENT', async () => {
+    const { credentialsPath } = makeCreds();
+
+    await expect(
+      runCreate(
+        {
+          profile: 'default',
+          output: 'json',
+          debug: false,
+          type: 'backend',
+          name: 'Guarded',
+          passwordFile: missing,
+        },
+        {
+          credentialsPath,
+          fetchImpl: (async () => {
+            throw new Error('should not hit network');
+          }) as unknown as typeof fetch,
+          stdout: () => {},
+          stderr: () => {},
+        },
+      ),
+    ).rejects.toMatchObject({
+      nextAction: expect.stringContaining('--password-file') as unknown as string,
+    });
+  });
+
+  it('still reads a password file that exists', async () => {
+    const { credentialsPath } = makeCreds();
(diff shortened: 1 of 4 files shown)


zeshi-du (approved):
Approving and merging. This is the right shape: one guarded reader in `src/lib/secret-file.ts` rather than a try/catch at each call site, mapped onto the same `VALIDATION_ERROR` envelope as the existing `readCodeFileGuarded`, with the directory case rejected up front (so EISDIR-vs-empty-read platform divergence can't leak through) and a UTF-8 BOM stripped for files written by PowerShell's default `Set-Content -Encoding utf8`. The rationale comments are doing real work — thank you for writing them.

Two follow-ups, neither blocking this merge:

1. @OkeyAmy — #283 covers the `--credential-file` / auto-auth flags. Please rebase onto this and call `readSecretFileGuarded()` instead of inlining the guard; this helper was written with exactly those flags as the intended next migration, and its own docstring says so. (Your Windows unit test is also failing now that CI has actually run — details on that PR.)

2. The size-cap question stays open and I'll record the decision on #58 rather than settle it in two different places: this PR deliberately omits a cap, #283 proposes `PAYLOAD_TOO_LARGE`. Whichever way it goes, it belongs in the shared helper so both flag families behave identically.

Sorry this sat 8 days fully green with no review — that was on us, not you.

PR #306, opened 2026-08-11
feat: add optional viewport field to plan files + --viewport CLI flag for responsive/mobile testing

### LINKED ISSUE #174: [Hackathon] Frontend runner uses a fixed desktop viewport - cannot test mobile/responsive layouts
## Summary

The frontend runner appears to execute at a **fixed desktop viewport width**, and there is no way (via a CLI flag or a plan step) to run a test at a mobile viewport. Responsive / mobile-only UI — e.g. a `md:hidden` bottom navigation bar — therefore can't be verified.

## Environment

- CLI: `testsprite` (latest)
- Test type: frontend (`--plan-from`)
- Target: live Next.js app using Tailwind responsive breakpoints

## What happened

I wrote a plan whose first step was "resize the browser to 390px width" to test a mobile bottom-nav. Across two runs:

- The runner ignored the resize step and instead appended `?mobile=1` to the URL (a no-op for the app).
- The desktop sidebar stayed visible; the `md:hidden` bottom nav never rendered.
- The mobile layout was never exercised, so the run couldn't assert on it.

## Expected

A way to set the viewport / device for a frontend run — for example a `--viewport <WxH>` or `--device <name>` flag on `test create` / `test run`, or a recognized "set viewport" plan step that the runner honors.

## Actual

No viewport / mobile / device / width / height option exists on any command (checked `testsprite --help` and `test create/run --help`, and the CLI source). The runner defaults to a desktop viewport with no override, so mobile-responsive assertions are untestable.

## Impact

Any app with responsive or mobile-specific UI can't be covered. We had to delete an otherwise-valid mobile bottom-nav test and document it as a runner limitation rather than hide a red result.

## Suggested fix

Add a viewport / device override (a flag, or a plan step the runner honors) so responsive layouts can be tested at defined widths.

---

Filed as part of TestSprite S3 Hackathon dogfooding (project: **Nora** — https://github.com/lxcario/Nora). Happy to provide the exact plan JSON and run IDs.

waterWang:
## Summary

Adds an optional `viewport` field to the plan file schema (`<width>x<height>`, e.g. `"390x844"`) and a `--viewport <WxH>` CLI flag to `test create --plan-from`. This lets users test responsive/mobile-only UI (e.g. `md:hidden` bottom nav) by telling the frontend browser runner what viewport to use before executing plan steps.

Closes #174

### `src/commands/test.ts`
- `CliPlanInput` — added optional `viewport?: string` field
- `assertPlanShape` — validates viewport format (`^\d+x\d+$`)
- `collectPlanIssues` — same viewport format check for `test lint`
- `runCreateFromPlan` — passes `viewport` through to `POST /tests` body; also supports `--viewport` CLI flag as an override of the plan file's viewport
- `test create` command — added `--viewport <WxH>` flag
- `CreateFromPlanOptions` — added `viewport?: string` for the override path
- `CreateFlagOpts` — added `viewport?: string` for the flag parsing

### `schemas/plan.schema.json`
- Added `viewport` property (string, pattern `^[1-9]\d*x[1-9]\d*$`)

### `DOCUMENTATION.md`
- Added `viewport` to the plan file field table

### `src/lib/plan-schema.spec.ts`
- Added test: validates valid viewport ("390x844") and rejects invalid ("abc")
- Added test: plan with viewport + all optional fields

## Testing

- `src/lib/plan-schema.spec.ts` — 17 tests pass (2 new): schema + validator both accept valid viewport, reject invalid
- `src/commands/test.test.ts` — 324 tests pass
- `src/commands/test.run.spec.ts` — 92 tests pass

### Plan file (JSON):
```json
{
  "projectId": "prj_abc123",
  "type": "frontend",
  "name": "Mobile bottom-nav test",
  "viewport": "390x844",
  "planSteps": [
    { "type": "action", "description": "Navigate to dashboard" },
    { "type": "assertion", "description": "Verify the mobile bottom navigation bar is visible" }
  ]
}
```

### CLI flag override:
```bash
testsprite test create --plan-from plan.json --viewport 390x844
```

When `--viewport` is supplied alongside `--plan-from`, it overrides the plan file's viewport value — useful for running the same plan at different breakpoints (e.g. mobile smoke pass on a desktop-authored plan).

## Summary by CodeRabbit

- **New Features**
  - Frontend test plans can now specify a browser viewport using `WIDTHxHEIGHT` format.
  - Added a `--viewport` option when creating tests from plans.
  - Command-line viewport values override the value defined in the plan.

- **Bug Fixes**
  - Added validation to reject invalid viewport dimensions and provide consistent plan validation results.

- **Documentation**
  - Documented viewport configuration and the desktop default used when no viewport is specified.


zeshi-du:
Thanks for building this, and CI is now approved (it had never run — our fault). Two separate things to say: one about the code, one about sequencing.

**Code:** `Lint & Format`, `Unit Tests` (Node 20 and 22) and `Coverage` all fail right now. There's also a mismatch worth fixing regardless of the outcome below — the viewport validation in the CLI and the shape documented for plan files don't agree on the accepted format, so a plan file that the schema would accept can be rejected by the flag and vice versa. Those need to be one definition.

**Sequencing, which is the bigger issue:** the backend does not currently accept a `viewport` field on plan steps or run triggers. Merging the CLI half alone would ship a flag that validates its input, sends it, and has no effect — a silent no-op is worse than a missing feature, because someone will write a mobile-layout test suite, watch it run at desktop width, and trust the pass.

This is the line `VISION.md` draws: *"Test execution, plan generation, browser automation ... live in the TestSprite platform, not in this CLI. The CLI surfaces them; it does not implement them."* A viewport is an execution parameter, so the platform has to accept and honour it first; then this PR is the surface for it and lands quickly.

So I'm not going to merge this yet, and I don't want to leave you guessing either. Concretely:
- I'm raising the platform-side change (accept and honour a viewport on FE runs) internally, referencing this PR and issue #174 as the demand signal — a working PR is a much better argument than a feature request, so this wasn't wasted.
- Once the backend accepts the field, this becomes a merge-on-green.
- If you want to keep it warm, fixing the lint/test failures and unifying the schema-vs-flag validation is useful work either way.

@lxcario's original report (#174) is what put this on the list; the desktop-only viewport is a genuine gap for anyone testing responsive layouts.

zeshi-du:
Status, so this isn't sitting on an open-ended pause: **this stays on hold, and there's now a dated decision point rather than an indefinite one.**

Nothing about my 08-13 read has changed — I re-verified it today rather than repeating it. The backend still accepts no `viewport` field anywhere: not on plan steps, not on the run-trigger path, on the portal route or the CLI one. And because unrecognized fields are stripped rather than rejected, merging the CLI half as-is would ship a flag that parses its input, validates it, sends it, and has it silently dropped. Someone would then write a mobile-layout suite, watch it run at desktop width, and have no way to tell why. A silent no-op is worse than a missing feature, which is why I can't take this one on its own.

What's new is that the platform side finally has a live owner and a date. There's an internal effort that just kicked off on mobile testing as a first-class test type, with a design review on **2026-08-22**. I've put the viewport question into that review explicitly, framed as its own decision rather than an implied one: native device testing and desktop-browser viewport emulation are different features that want the same backend plumbing, and somebody has to say which one is in scope — because right now nobody has, and the default outcome of not deciding is that your PR sits here forever.

So: I'll come back to this thread after the 22nd with a real answer — either "the backend is taking a viewport field, rebase and this lands as the CLI surface for it," or "viewport emulation is out of scope in favour of device testing, here's why," which would be a decline with a reason instead of a pause.

One thing worth fixing regardless of that outcome, since it's a real defect independent of the sequencing: the viewport validation in the CLI and the shape documented for plan files don't agree on the accepted format, so a plan file the documented schema would accept can be rejected by the flag and vice versa. Those need to be one definition. `Lint & Format`, `Coverage` and two `Unit Tests` legs are also red on the current head.

To be explicit about what I'm not saying: the feature is wanted and the need behind it is real and well-documented internally. This isn't a rejection, and it isn't your code being wrong about anything except the schema mismatch above.

zeshi-du:
An update, and a straight answer about where this lands — you've waited a long time for one.

**The direction was adopted. It's being built on the platform side, not in the CLI.** Per-case and per-project viewport pinning is in flight in the backend and the portal right now. That's what makes this patch the wrong shape rather than the wrong idea: once the backend owns a viewport field, the CLI's job is to send *that* field, not to define a competing one in the plan file.

Two concrete reasons not to merge it today:

1. `viewport` appears nowhere in `main`, so this would be the CLI inventing a plan-file key that nothing downstream reads.
2. The backend half is not in production yet. The flag would validate locally and then quietly do nothing on a real run — which is worse than not having the flag, because it looks like it worked.

There's also a known gap on the platform side where a pinned viewport is ignored on the replay path. They're still closing that, which is another reason not to freeze a CLI surface against it yet.

**So: I'm closing this PR and un-assigning #174, but keeping #174 open as the tracking issue.** This is not a judgement on the patch — you identified a real limitation and it is being fixed because of reports like yours.

If you'd like to take the CLI half once the backend field is live, say so on #174 and it's yours. I'll comment there when the wire lands rather than leaving you to poll it.

PR #308, opened 2026-08-13
fix(security): baseline the fs-lint backlog + scope the gate to changed lines

fyZhang66:
## Problem

The `Security` workflow's `ESLint Security (changed files)` job failed on the **v0.6.0 release push** with `141 problems (128 errors)` — all `security/detect-non-literal-fs-filename` in `src/commands/test.ts` (+ `test.test.ts`).

None of these were introduced by v0.6.0. The job lints **changed FILES**, and ESLint lints a file **whole** — so touching the ~11k-line `test.ts` surfaced its entire pre-existing backlog of `fs.X(<variable path>)` calls, on lines the release never touched. As the workflow header already notes, a full-tree run yields ~380 such findings, almost all ordinary local-path fs calls a CLI makes routinely — not a real vulnerability signal for this codebase shape. So "changed files only" doesn't actually bound the noise once a large legacy file is touched.

## Fix — two layers, no rule severity lowered, no part of the tree exempted

- **(C) Baseline** — commit `eslint-suppressions.json` (ESLint's native bulk suppressions, auto-read from repo root) recording today's backlog. Pre-existing findings are no longer re-reported; a **genuinely new** fs call still surfaces because the per-file count exceeds the baseline.
- **(A) Changed-line scope** — `.github/scripts/filter-changed-line-findings.mjs` keeps only error findings on lines this diff **actually added or modified**, so touching a legacy file never fails on its untouched lines. The lint step now emits `--format json` and runs with `--pass-on-unpruned-suppressions`, so *fixing* a baselined finding in a changed file doesn't fail the job on a now-stale suppression entry.

Net: every `security/*` rule keeps its designed severity, new/changed code is still held to it, and the existing tree is neither re-linted from scratch nor exempted. Retire baseline entries as they get fixed with `eslint --prune-suppressions` in a maintenance pass (never regenerate it to hide new findings).

## Verified locally (against the real commits)

- **v0.6.0 release push** (`b35dbaeb`→`aa517e7c`, 16 changed src TS files): ESLint exit 0, filter reports no new findings → **the release would go green**.
- **A genuinely new `writeFileSync(<var>)`**: surfaces past the baseline and the changed-line filter fails the job, pointing at the exact `file:line`.
- Fatal ESLint errors (config/crash, exit ≥ 2) still hard-fail — only lint findings are deferred to the changed-line decision.
- `eslint-suppressions.json` records only the `detect-non-literal-fs-filename` error rule (warnings are unaffected and still print). Added to `.prettierignore` since it is tool-generated.

## Note

This PR touches only `.github/**` + config, no `src/**/*.ts`, so its own security job sees zero changed source files and passes cleanly.

## Summary by CodeRabbit

* **Security**
  * Security checks now focus on newly added or modified code, making actionable issues easier to identify.
  * Existing, reviewed findings are tracked through suppression baselines while new high-severity issues continue to fail validation.
* **Chores**
  * Improved linting reliability and handling of files with special characters.
  * Added documentation for managing security suppressions and regenerating the baseline.

### DIFF
--- a/.github/scripts/filter-changed-line-findings.mjs
+++ b/.github/scripts/filter-changed-line-findings.mjs
@@ -1,0 +1,93 @@
+/**
+ * Fail the security lint only on ERROR-severity findings that land on lines
+ * THIS change actually added or modified.
+ *
+ * This is the "changed-line" half of the gate. The other half — the
+ * committed `eslint-suppressions.json` baseline — has already removed the
+ * pre-existing backlog from the ESLint report before it reaches this script,
+ * so what remains is genuinely new/excess findings; this step additionally
+ * narrows them to the diff's own lines so a large legacy file (e.g.
+ * `src/commands/test.ts`) does not fail a release just for being touched.
+ *
+ * Input:  argv[2] = path to an ESLint JSON report (already baseline-filtered)
+ *         env RESOLVED_BASE = base commit to diff against ('' = no base)
+ *         env HEAD_SHA      = head commit (default 'HEAD')
+ * Exit:   1 if any error finding falls on an added/changed line, else 0.
+ */
+import { readFileSync } from 'node:fs';
+import { execFileSync } from 'node:child_process';
+
+const reportPath = process.argv[2];
+if (!reportPath) {
+  console.error('usage: filter-changed-line-findings.mjs <eslint-report.json>');
+  process.exit(2);
+}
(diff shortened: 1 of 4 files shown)


fyZhang66:
Wrong target — this belongs in the source repo, not this mirror. Moving the fix there.

PR #309, opened 2026-08-14
fix(ticker): respect terminal width during progress updates

### LINKED ISSUE #72: [Hackathon] Width-aware progress ticker

interferon0

The test run --wait / test wait progress ticker prints a single in-place line. On a narrow terminal or with a long run id the line soft-wraps to a second row, and the next \x1b[2K\r clears only one of the two rows, corrupting the redraw (ghost lines). This clips each frame to the current terminal width, re-read on every poll so it also follows a live resize. It stays completely silent on non-TTY (CI) and under --output json, so machine output is unchanged.

Approach: width-aware truncation in src/lib/ticker.ts (read process.stderr.columns ?? 80, clip to width-1 to avoid last-cell auto-wrap, append a 1-column ellipsis). No new dependencies (ticker content is ASCII, so .length is an exact column count). Adds 10 unit tests (truncation, simulated resize, 80-col fallback, non-TTY silence). Full gate green: tsc, eslint, prettier, 1703 tests, coverage 98% on ticker.ts. PR to follow once assigned.

### Confirmations

guaguagf:
## Problem

The in-place stderr ticker can soft-wrap on narrow terminals or with long run IDs. The next `ESC[2K\r` clears only one row, leaving ghost output behind.

## Root cause

Ticker frames were written without considering `process.stderr.columns` or terminal display-column width.

## Fix

- Read the current stderr terminal width for every frame, so live resizes are respected.
- Clip updates and final lines to `columns - 1` and append a one-column ellipsis when truncated.
- Measure terminal display columns for wide glyphs, emoji, combining marks, variation selectors, and controls.
- Truncate only at grapheme-cluster boundaries via `Intl.Segmenter`, so user-perceived characters are never split.
- Fall back to 80 columns when the width is unavailable or invalid.
- Preserve non-TTY silence and ANSI-free `NO_COLOR` output.
- Add no runtime dependencies.

## Tests

- 27/27 focused ticker tests passed, including truncation, live resize, 80-column fallback, final output, `NO_COLOR`, CJK width, emoji graphemes, and combining sequences.
- `npm test` — passed
- `npm run test:e2e` — 66 passed, 1 skipped
- `npm run lint` — passed
- `npm run typecheck` — passed
- `npm run format:check` — passed
- `npm run build` — passed

## Type of change

## Checklist

- [x] Commits follow Conventional Commits.
- [x] Lint, formatting, typecheck, unit tests, and coverage gate pass.
- [x] New behavior is covered by unit tests.
- [x] No secrets, credentials, or personal data are included.

## Risk

Low. The change is isolated to ticker rendering, adds no dependencies, and leaves non-TTY behavior unchanged.

Closes #72

## Bug Fixes

* Ticker output now fits within the available terminal width.
* Long updates and final lines are truncated at grapheme boundaries with an ellipsis.
* Improved display-width handling for wide characters, emoji, and combining marks.
* Added a fallback for terminals without a reported width.
* Width-limited output now works correctly with or without color formatting.
* Zero-width terminals produce empty ticker output instead of overflowing.

### DIFF
--- a/src/lib/ticker.spec.ts
+++ b/src/lib/ticker.spec.ts
@@ -276,6 +276,57 @@ describe('createTicker — terminal width', () => {
       expect(lines[0]).not.toContain('\x1b[2K');
     });
   });
+
+  it('measures wide glyphs by display columns', () => {
+    withStderrColumns(32, () => {
+      const raw: string[] = [];
+      const ticker = createTicker(
+        () => {},
+        true,
+        text => raw.push(text),
+        false,
+      );
+
+      ticker.update('界界界界');
+
+      const rendered = raw[0]!.slice('\x1b[2K\r'.length);
+      expect(rendered.slice(rendered.indexOf(' ') + 1)).toBe('界界…');
+    });
+  });
+
+  it('does not split an emoji grapheme cluster', () => {
+    withStderrColumns(29, () => {
+      const raw: string[] = [];
+      const ticker = createTicker(
+        () => {},
+        true,
+        text => raw.push(text),
+        false,
+      );
+
+      ticker.update('👨‍👩‍👧‍👦xx');
+
+      const rendered = raw[0]!.slice('\x1b[2K\r'.length);
+      expect(rendered.slice(rendered.indexOf(' ') + 1)).toBe('👨‍👩‍👧‍👦…');
+    });
+  });
+
+  it('keeps combining sequences intact and counts them as one column', () => {
+    withStderrColumns(29, () => {
+      const raw: string[] = [];
+      const ticker = createTicker(
+        () => {},
+        true,
+        text => raw.push(text),
+        false,
+      );
+
+      ticker.update('e\u0301e\u0301xx');
+
+      const rendered = raw[0]!.slice('\x1b[2K\r'.length);
+      expect(rendered.slice(rendered.indexOf(' ') + 1)).toBe('e\u0301e\u0301…');
+    });
+  });
 });
 
 describe('createTicker — default isTTY detection', () => {
--- a/src/lib/ticker.spec.ts
+++ b/src/lib/ticker.spec.ts
@@ -55,6 +55,24 @@ function stripTickerTimestamp(raw: string): string {
   return prefix + afterEsc.slice(spaceIdx + 1);
 }
 
+function withStderrColumns(columns: number | undefined, run: () => void): void {
+  const originalColumns = Object.getOwnPropertyDescriptor(process.stderr, 'columns');
+  Object.defineProperty(process.stderr, 'columns', {
+    configurable: true,
+    value: columns,
+  });
+
+  try {
+    run();
+  } finally {
+    if (originalColumns === undefined) {
+      Reflect.deleteProperty(process.stderr, 'columns');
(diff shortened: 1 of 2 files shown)


guaguagf:
@coderabbitai review

zeshi-du:
This sat without a human reply for five days — that's on us, not on the work. Sorry for the silence.

This is a genuinely good fix. `Intl.Segmenter` for grapheme-cluster boundaries, a real East-Asian-Width table for fullwidth code points, and explicit handling for emoji/ZWJ sequences, variation selectors, and combining marks is more careful than most terminal-width code I see. Re-reading `process.stderr.columns` on every frame instead of caching it means live resize just works without a `SIGWINCH` listener, and the 80-column fallback covers the "columns is unavailable" case cleanly. Zero new dependencies for all of that. I also checked that the `!tty` early return is untouched, so piped/non-TTY/CI output is exactly what it was before this PR.

One thing I'd like changed before merge. `fitToTerminal` is applied in both branches — the `NO_COLOR` branch (ticker.ts:179 and :185, writing through `stderrWrite`) and the ANSI branch (:197 and :203, writing through `rawWrite` with `\x1b[2K\r`). But `stderrWrite` appends a newline: that branch always scrolls, one line per call, and never does the in-place `ESC[2K\r` rewrite. The ghost-row bug in #72 is specifically a soft-wrap-plus-overwrite problem — it can't happen on a path that never overwrites. So clipping the `NO_COLOR` branch buys no correctness there and does cost something real: the tail of every long line, for anyone running with `NO_COLOR` set on a real TTY — a screen reader or a high-contrast terminal setup, for instance, since `!tty` already makes piped/redirected output completely silent regardless of `NO_COLOR`. Can you confine the clipping to the in-place (ANSI) branch and leave `NO_COLOR` output unclipped? The `clips NO_COLOR output without adding ANSI sequences` spec would need to flip to asserting pass-through instead.

Two smaller things, non-blocking:
- The fullwidth table is a solid approximation but not exhaustive against the full East-Asian-Width spec — fine as-is, just flagging it as a known approximation.
- `terminalWidth` doesn't discount ANSI sequences embedded inside a caller-supplied `line`. I checked every current call site and none of them pass colorized text today, so this is latent, not a bug — just something to keep in mind if a future caller ever does.

Separately: fork CI never ran on this PR — the approval gate needed a maintainer click and nobody clicked it for five days. I've approved it now and the runs have completed (head `6832e83`). Everything is green except `Dependency Review`, which fails on every PR in this repo right now because our org-level dependency graph is disabled — that's on our infrastructure, not this diff. Ignore it.

One more thing worth knowing for next time: that approval gate re-arms on every new push from a first-time contributor, and the nudge bot that pings on CI failure only fires on a *completed* run — a run that never starts produces no notification at all, so your checks tab would just look empty with no explanation. If a future push here (or on #310) sits with no checks running, ping rather than assume something's wrong on your end.

zeshi-du:
Sorry for the silence since 2026-08-19.

Status check today: this still merges cleanly against current `main` (v0.11.0) despite four releases landing in between — nothing needed from you. Picking it back up; the remaining questions are ours to answer, not yours.

PR #310, opened 2026-08-15
feat(test): warn on conditional frontend assertions

### LINKED ISSUE #175: [Hackathon] Runs return verdict: blocked on verbose/multi-branch assertions even when the page is correct
## Summary

Frontend runs intermittently finish with `verdict: blocked` / `executionStatus: completed` **even when the page under test rendered correctly**. The trigger is the *shape* of the plan/assertion, not a product defect: verbose "either A or B … and not broken" OR-assertions, or plans that navigate many items sequentially, exhaust the agent's step/runway budget before it emits a clean pass/fail verdict.

## Environment

- CLI: `testsprite` (latest)
- Test type: frontend (`--plan-from`)

## Reproduction pattern

1. Author a plan whose final assertion is a two-branch OR, e.g. *"verify either an interactive graph canvas OR a clear empty-state message, rather than an empty or broken container."*
2. Run against a page that clearly satisfies one branch (heading + empty-state button visible).
3. The agent re-checks both branches; its own run summary states the page is correct ("PASS"), but the verdict comes back `blocked` because the budget ran out before a decisive verdict was emitted.

The same failure mode reproduces with plans that walk many nav items in a single run (e.g. 11 sidebar links sequentially — every executed step passed, but the run blocked before finishing).

## Expected

- Verbose / conditional assertions shouldn't consume enough budget to block a verdict on a correct page, **or**
- The runner should surface a clearer signal that the block was budget/plan-shape driven (distinct from an infra block), **and/or**
- The practical assertion-complexity ceiling should be documented so authors know the limit up front.

## Actual

`verdict: blocked` with no product defect. The only reliable fix we found was to rewrite each assertion to be single and decisive ("verify heading X and button Y") — a plan-authoring workaround rather than a runner behavior.

## Impact

Correct pages report as blocked, which is easy to misread as a product failure. Authors have to learn by trial that FE assertions must be minimal — this isn't documented anywhere.

## Suggested fix

Either raise (and clearly report) the per-run step budget, distinguish "blocked (budget / plan-shape)" from "blocked (infra)", or document an assertion-complexity guideline for frontend plans.

---

Filed as part of TestSprite S3 Hackathon dogfooding (project: **Nora** — https://github.com/lxcario/Nora). Happy to share the exact plan JSON, the `test plan put` fix, and run IDs for the blocked→green arcs.

guaguagf:
## What does this PR do?

Frontend plans with conditional or multi-branch assertions can exhaust the browser agent's run budget and return `blocked` even when the page is correct. Today, `test lint` only checks structural validity, so these valid-but-costly plans pass preflight without guidance.

The offline linter did not inspect assertion wording for plan-shape patterns such as `either A or B`, `if A then B`, `unless`, or `whether`.

This PR:
- adds a narrow, non-fatal assertion-complexity heuristic to `test lint`
- reports warnings only for frontend `assertion` steps
- keeps text-mode advisories on stderr while JSON warnings remain structured on stdout
- preserves exit 0 and existing JSON/text output when no warning exists
- documents the single-decisive-assertion guideline
- adds positive and negative regression coverage

## Related issue

Closes #175

## Type of change

- [ ] Documentation only
- [ ] Build / CI / chore

## Checklist

- [x] Commits follow [Conventional Commits](https://www.conventionalcommits.org).
- [x] `npm test` passes and coverage stays at or above the 80% gate.
- [x] User-facing changes are reflected in the CLI skill documentation.

## Validation

- Unit: 66 files passed; 2,339 tests passed, 2 skipped
- E2E: 4 files passed; 66 tests passed, 1 skipped
- ESLint passed
- Prettier passed
- TypeScript typecheck passed
- Build passed

## Notes for reviewers

Risk is low. The heuristic is local and advisory-only: it does not reject structurally valid plans or change run dispatch behavior. Actions and single-outcome assertions remain unchanged.

### DIFF
--- a/docs/cli-v1-agent-install/skill-template.md
+++ b/docs/cli-v1-agent-install/skill-template.md
@@ -173,6 +173,10 @@ blue 'Submit' button"`. The agent does its own semantic match; a guessed label
 - **1–2 assertions, at the end, on content existence** rather than UI copy /
   formatting. (Quoting a literal you submitted earlier in the same plan, to verify
   it round-trips, is fine.)
+- **Keep each assertion single and decisive.** Avoid conditional or multi-branch
+  wording such as "either A or B", "if A then B", `unless`, or `whether` — it can
+  exhaust the frontend run budget before a verdict is emitted. `test lint` warns
+  on these patterns before the plan consumes run credits.
 - **Assertion targets name the specific page region** (panel / tab / output area).
   Otherwise the agent settles for "some element with that text is visible
   anywhere," which passes for wrong reasons.
--- a/src/commands/test.lint-warning.spec.ts
+++ b/src/commands/test.lint-warning.spec.ts
@@ -1,0 +1,67 @@
+import { mkdtempSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
+import { runLint } from './test.js';
+
+describe('runLint assertion-complexity warnings', () => {
+  it('warns when a frontend assertion contains conditional or multi-branch wording', async () => {
+    const dir = mkdtempSync(join(tmpdir(), 'cli-lint-conditional-'));
+    const file = join(dir, 'plan.json');
+    writeFileSync(
+      file,
+      JSON.stringify({
+        projectId: 'project_alice',
+        type: 'frontend',
+        name: 'Knowledge Web renders',
+        planSteps: [
+          {
+            type: 'assertion',
+            description: 'Verify either an interactive graph canvas or a clear empty-state message',
+          },
+        ],
+      }),
+      'utf8',
+    );
+
+    const report = await runLint(
+      { profile: 'default', output: 'json', debug: false, planFrom: file },
+      { stdout: () => undefined },
+    );
+
+    expect(report).toMatchObject({ checked: 1, valid: 1, issues: [] });
+    expect(report.warnings).toEqual([
+      expect.objectContaining({
+        field: 'planSteps[0].description',
+        reason: expect.stringContaining('single, decisive assertion'),
+      }),
+    ]);
+  });
+
+  it('does not flag actions or single-outcome assertions', async () => {
+    const dir = mkdtempSync(join(tmpdir(), 'cli-lint-decisive-'));
+    const file = join(dir, 'plan.json');
+    writeFileSync(
+      file,
+      JSON.stringify({
+        projectId: 'project_alice',
+        type: 'frontend',
+        name: 'Checkout works',
+        planSteps: [
+          { type: 'action', description: 'Open the cart or return to the catalog' },
+          { type: 'assertion', description: 'Verify the order total is visible' },
+        ],
+      }),
+      'utf8',
+    );
+
+    const report = await runLint(
+      { profile: 'default', output: 'json', debug: false, planFrom: file },
+      { stdout: () => undefined },
+    );
+
+    expect(report).toEqual({ checked: 1, valid: 1, issues: [] });
+  });
+});
--- a/src/commands/test.lint-warning.spec.ts
+++ b/src/commands/test.lint-warning.spec.ts
@@ -40,6 +40,38 @@ describe('runLint assertion-complexity warnings', () => {
     ]);
   });
 
+  it('keeps text warnings on stderr and stdout machine-safe', async () => {
+    const dir = mkdtempSync(join(tmpdir(), 'cli-lint-warning-streams-'));
+    const file = join(dir, 'plan.json');
+    writeFileSync(
+      file,
+      JSON.stringify({
+        projectId: 'project_alice',
+        type: 'frontend',
+        name: 'Knowledge Web renders',
+        planSteps: [
+          {
+            type: 'assertion',
+            description: 'Verify either an interactive graph canvas or a clear empty-state message',
+          },
+        ],
+      }),
+      'utf8',
+    );
+    const stdout: string[] = [];
+    const stderr: string[] = [];
+
+    await runLint(
+      { profile: 'default', output: 'text', debug: false, planFrom: file },
+      { stdout: line => stdout.push(line), stderr: line => stderr.push(line) },
+    );
+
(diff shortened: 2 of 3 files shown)


zeshi-du:
Like #309, this went four days without a human reply, which is on us. Sorry for that.

The scoping here is exactly right, and I want to say so explicitly because it's easy to get wrong: #175 was split at triage into an engine-side half (budget-exhaustion verdicts, out of scope for this repo, tracked internally) and a CLI-side half — (a) document the assertion-complexity ceiling, (b) have `test lint` warn before a run is dispatched and paid for. This PR implements exactly (a) and (b) and nothing more. The design is right too: warnings are non-fatal, `issues`/`valid`/`checked` and the exit code are untouched, text mode keeps warnings on stderr so stdout stays machine-parseable, and JSON mode adds `warnings` only when it's non-empty rather than always emitting an empty array.

The one thing I want changed: the heuristic is wider than it needs to be.

```
/\b(?:either|or|otherwise|unless|whether)\b|\bif\b[^.!?]*\bthen\b/i
```

A bare `\bor\b` matches any assertion that contains the word "or" for any reason, and on-screen UI copy is full of it. Two concrete cases that would be wrongly flagged, both single, decisive, non-branching assertions:

- `Verify the button reads "Sign in or create account"`
- `Verify the modal title is "Import or export data"`

I ran the regex against both — it fires on each one. The blast radius of a false positive is small today (a stderr line, no gate on the run), and the bare `or` is an easy thing to reach for in a heuristic — the guidance's own examples ("either A or B") lean on that word too — but I'd still rather not warn on assertions that are correctly quoting product copy. The PR's own negative test only covers an *action* step containing "or"; the gap is specifically a decisive *assertion* that quotes text containing "or." Could you tighten this to one of:
- require an actual branching context rather than a bare `or` — e.g. `either ... or`, `or otherwise`, or two full noun phrases joined by `or` — or
- skip text inside quotation marks, since that's almost always literal copy being asserted on, not branching logic

Either approach works for me; pick whichever is less awkward to implement, or do both.

Second item: the new guidance only landed in `docs/cli-v1-agent-install/skill-template.md`. The reshaped issue's whole point was that this should stop being knowledge people acquire by trial — and right now the only humans who see it are the ones reading that install-skill template. `DOCUMENTATION.md` has a "Plan file format" section with the `planSteps` field table, and it already carries an authoring-guidance callout right below that table (the `{{...}}`-placeholder advisory) — that's the precedent for where this kind of guidance belongs. Can you add the single-decisive-assertion guidance there too, so it's discoverable from the user-facing docs and not just the agent-install template?

Also, factually, so you don't spend time on it: CodeRabbit's "Merge Risk: Moderate" banner on this PR is stale — it claims "the new regression test currently passes an extra dependency argument, causing the test file to fail TypeScript compilation." I checked it against your head commit directly: `runLint` is declared `(opts: LintOptions, deps: TestDeps = {})` (test.ts:4821), `TestDeps` declares both `stdout?` and `stderr?` (test.ts:446-447), and every `runLint(...)` call in your new spec passes exactly two arguments matching that shape. Typecheck and both Node builds are green at your head SHA (`ffbb942`). There's no compile failure — don't chase that one.

Same fork-CI note as #309: the approval gate never got clicked here either. I've approved it now (head `ffbb942`), so the real checks have finally run. Everything that matters is green — `Unit Tests` on Node 20, Node 22 and Windows, `Coverage`, `Typecheck`, `Lint & Format`, both `Build`s and `Local E2E Tests`.

Two red checks, and they are not the same story — I want to be precise rather than tell you to ignore both:

- **`Dependency Review` — ours, ignore it.** It fails on every PR in this repo because dependency graph is disabled at the org level. Nothing to do with your change.
- **`ESLint Security (changed files)` — mostly ours, but three findings are genuinely yours.** The check lints a changed *file* whole, so touching `src/commands/test.ts` re-surfaces its entire pre-existing `security/detect-non-literal-fs-filename` backlog on lines you never went near; that half is ours and the baseline fix already exists in our source repo, shipping next release. But the log also flags your new file:

  ```
  src/commands/test.lint-warning.spec.ts
    13:5  error  Found writeFileSync from package "node:fs" with non literal argument at index 0
    46:5  error  Found writeFileSync from package "node:fs" with non literal argument at index 0
    78:5  error  Found writeFileSync from package "node:fs" with non literal argument at index 0
  ```

  These are the three `writeFileSync(file, …)` calls where `file` is built from `mkdtempSync`. They're safe, and the repo already has a pattern for saying so at the call site — a targeted disable with the justification spelled out, e.g.:

  ```ts
  // eslint-disable-next-line security/detect-non-literal-fs-filename -- plan fixture written into this test's own mkdtempSync-created temp dir, never user input.
  writeFileSync(file, JSON.stringify(plan), 'utf8');
  ```

  One warning that will save you a confusing round trip: **do not run `npm run lint:fix` after adding those comments.** The security rules only load under the dedicated security config, so the main ESLint pass sees the directive as unused and `--fix` deletes it — silently, while `npm run lint` still passes. Add the three comments by hand and leave them alone. That's a sharp edge on our side, not yours.

And the same heads-up: this gate re-arms on every push from a first-time contributor, and a run that never starts triggers no failure notification — your checks tab will just look empty. If that happens again on a future push, ping instead of waiting on it.

zeshi-du:
Same apology as on #309 — this has been sitting since 2026-08-19.

It still merges cleanly against current `main` (v0.11.0), so nothing is needed from you right now. The open question from that review is a product one — whether a conditional frontend assertion should warn or be rejected at authoring time — and that's ours to decide, not something to leave hanging on you. I'll come back with an answer rather than another round of silence.

zeshi-du (changes requested):
Hi @guaguagf — thanks for this; the phrasing heuristic reads well. Two things first:

- The warning only fires for `--plan-from` object input. A `--steps` JSON array containing the same assertion wording (e.g. `{"type":"assertion","description":"either A or B"}`) skips it entirely, since `collectAssertionComplexityWarnings` short-circuits on arrays. Please add array-input coverage so both `--steps` and `--plan-from` get the same treatment.
- The Security workflow's ESLint check (`security/detect-non-literal-fs-filename`) flags the new fixture writes in `src/commands/test.lint-warning.spec.ts` (~lines 13, 46, 78); please add the standard suppressions there.

Happy to take another look once those are addressed.

PR #319, opened 2026-08-21
fix(cron): stop describing a month-restricted schedule as daily

### LINKED ISSUE #318: bug: schedule advisory calls a month-restricted cron "daily", contradicting the frequency beside it
While reading `src/lib/cron.ts` I noticed `describeCron` destructures the five cron fields but skips the month:

```ts
const [minute, hour, dayOfMonth, , dayOfWeek] = f;
```

So a schedule that is restricted to certain months gets described by its finer fields alone. `runsPerMonth` handles the month correctly (it divides by `months.size / MONTHS_PER_YEAR`), which puts the two halves of the advisory in direct conflict.

**Reproduction**

```
$ testsprite schedule create --cron "0 3 * 6 *" ...
This schedule will run ~3 time(s)/month (daily at 03:00). Each run is a full test run.
```

`0 3 * 6 *` runs daily at 03:00 **during June only**. "~3 time(s)/month" and "daily at 03:00" cannot both be true, and they are printed in the same sentence. This is the user-facing path — `formatScheduleFrequencyAdvisory` is called from `src/commands/schedule.ts:245` (create) and `:347` (update).

Confirmed against `main` (40d18d2):

| cron | actual meaning | `describeCron` |
|---|---|---|
| `0 3 * 6 *` | daily in June | `daily at 03:00` |
| `0 3 * 1,7 *` | daily in Jan & Jul | `daily at 03:00` |
| `0 3 * */6 *` | daily in Jan & Jul | `daily at 03:00` |
| `30 9 1 3 *` | 1 March, yearly | `monthly on day 1 at 09:30` |

The existing tests miss it because `describeCron` is never exercised with a non-wildcard month, and the one advisory test that uses one (`0 3 1 1 *`) asserts only the frequency half, not the description.

**Suggested fix**

None of the phrasings `describeCron` can produce is able to carry "only in these months", so a restricted month should fall back to the raw expression — the same thing the function already does for every other shape it cannot state exactly, and what its docstring asks for: *"a wrong description is worse than the raw expression"*.

I have the fix and regression tests ready (verified failing on `main` and passing with the change) and would like to pick this up — `/assign`.

naufalfx805-source:
Closes #318

`describeCron` destructured the five cron fields but skipped the month, so a schedule restricted to certain months was described by its finer fields alone. `0 3 * 6 *` — daily at 03:00, but only in June — came back as `daily at 03:00`.

That string is printed by `formatScheduleFrequencyAdvisory` beside a frequency `runsPerMonth` computes correctly, so `schedule create` (`src/commands/schedule.ts:245`) and `schedule update` (`:347`) rendered a sentence that contradicts itself:

```
This schedule will run ~3 time(s)/month (daily at 03:00).
```

**Fix.** None of the phrasings `describeCron` produces can carry "only in these months", so a restricted month now falls back to the raw expression — the same thing the function already does for every other shape it cannot state exactly, and what its own docstring asks for: *"a wrong description is worse than the raw expression"*. `runsPerMonth` is untouched; it already handled the month correctly.

**Tests.** Two added, both verified failing on `main` and passing with the change:

- `describeCron > returns the raw expression when the month is restricted` — covers `0 3 * 6 *`, `0 3 * 1,7 *`, `0 3 * */6 *`, `0 3 1 3 *`, `0 3 * 6 1`. Without the fix: `expected 'daily at 03:00' to be '0 3 * 6 *'`.
- `formatScheduleFrequencyAdvisory > does not call a month-restricted schedule daily` — asserts the advisory keeps `~3 time(s)/month` and stops saying `daily`.

`npx vitest run src/lib/cron.test.ts` → 24 passed. `src/commands/schedule` → 66 passed.

The gap in the existing suite was that `describeCron` was never exercised with a non-wildcard month, and the one advisory test that uses one (`0 3 1 1 *`) asserts only the frequency half, not the description.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cron schedule descriptions for expressions restricted to specific months.
  * These schedules now display the original cron expression instead of an inaccurate daily, weekly, or monthly description.
  * Preserved reduced monthly frequency information in schedule guidance.

### DIFF
--- a/src/lib/cron.test.ts
+++ b/src/lib/cron.test.ts
@@ -107,6 +107,15 @@ describe('describeCron', () => {
     }
   });
 
+  it('returns the raw expression when the month is restricted', () => {
+    // Every phrasing describeCron can produce names a cadence, and none of them
+    // can carry "only in these months". Saying "daily at 03:00" for a June-only
+    // cron contradicts the ~3 times/month printed next to it by the advisory.
+    for (const cron of ['0 3 * 6 *', '0 3 * 1,7 *', '0 3 * */6 *', '0 3 1 3 *', '0 3 * 6 1']) {
+      expect(describeCron(cron), cron).toBe(cron);
+    }
+  });
+
   it('returns the raw expression when the field count is not 5', () => {
     // A 6-field Quartz cron shifts every index, so reading f[1] as the hour
     // would describe a time the schedule never runs at.
@@ -143,6 +152,16 @@ describe('formatScheduleFrequencyAdvisory', () => {
     const out = formatScheduleFrequencyAdvisory('0 3 1 1 *');
     expect(out).toContain('less than once a month');
     expect(out).not.toContain('~0 time');
+  });
+
+  it('does not call a month-restricted schedule daily', () => {
+    // Regression: describeCron used to skip the month field entirely, so this
+    // rendered as "~3 time(s)/month (daily at 03:00)" -- two claims that cannot
+    // both be true, in one sentence.
+    const out = formatScheduleFrequencyAdvisory('0 3 * 6 *');
+    expect(out).toContain('~3 time(s)/month');
+    expect(out).not.toContain('daily');
+    expect(out).toContain('0 3 * 6 *');
   });
 
   it('still advises, without a frequency, for an expression it cannot read', () => {
--- a/src/lib/cron.ts
+++ b/src/lib/cron.ts
@@ -145,9 +145,15 @@ export function runsPerMonth(cron: string): number | null {
 export function describeCron(cron: string): string {
   const f = fields(cron);
   if (f.length !== FIELD_COUNT) return cron.trim();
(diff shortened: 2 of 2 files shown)


naufalfx805-source:

zeshi-du (approved):
Verified against current `main` (v0.11.0): merges cleanly, full CI green, and the fix is the right shape.

`describeCron` was destructuring past the month field (`const [minute, hour, dayOfMonth, , dayOfWeek]`), so `0 3 * 6 *` rendered as "daily at 03:00" while `runsPerMonth` — which *does* divide by the month count — printed `~3 time(s)/month` in the same sentence. Two claims that cannot both be true. Returning the raw expression is the correct call: none of the phrasings `describeCron` can produce is able to carry "only in these months", so there is nothing accurate to say.

Both regression tests are load-bearing — the second one asserts `not.toContain('daily')` against the exact advisory string that was wrong.

Sorry this sat for three and a half weeks. Thanks for the clean report-plus-fix pair.

PR #322, opened 2026-08-27
fix(project): cap --password-file size before reading it (#58)

### LINKED ISSUE #58: Unbounded Pagination and Arbitrary File Read via --password-file

naufalfx805-source:
## Summary

`--password-file` reads whatever path it is given, whole, into memory, and POSTs it to the
configured API server as `body.password` — with no size advisory and no confirmation. Point
it at a log, a database dump, or a binary by mistake and the typo becomes an upload.

This adds a 64 KB ceiling, checked on `statSync` **before** the read, so an obvious mistake
fails fast with `PAYLOAD_TOO_LARGE` (exit 5) without pulling the file into memory at all and
without spending a round-trip.

Closes #58.

## Why this doesn't contradict #302

The comment #302 left in `src/lib/secret-file.ts` says the cap was *"deliberately not carried
over: secrets are small, and a size ceiling would be a behaviour change on a shipped flag
rather than part of fixing the crash."* That was the right call for #302, which was scoped to
turning an unhandled Node exception into a typed envelope. This PR is the deferred follow-up
that argues the cap on its own merits, and it updates that comment accordingly.

## Sizing

64 KB is roughly 20x headroom over the largest secret these flags legitimately carry — a
4096-bit RSA private key in PEM is ~3.2 KB, a GCP service-account JSON key ~2.4 KB. Unlike
`MAX_INLINE_CODE_BYTES` it tracks no backend limit; the server still rejects oversize
credentials on its own terms. This is a client-side pre-flight, not a mirror of a server rule.

## Testing

- 104/104 pass across `src/lib/secret-file.test.ts` and `src/commands/project.test.ts`,
  including a file sitting exactly on the cap (accepted) and one byte over (rejected).
- `tsc --noEmit` clean; ESLint Security clean.

## Note

Rebased onto `main` at v0.7.0. The already-merged #302 commit drops out of the range
automatically, leaving just the size-cap commit. The two new test helpers write into their
own `mkdtempSync` directories and carry targeted
`eslint-disable-next-line security/detect-non-literal-fs-filename` justifications, matching
the existing idiom in `secret-file.ts`, so the security suppressions baseline is unchanged.

### DIFF
--- a/src/commands/project.test.ts
+++ b/src/commands/project.test.ts
@@ -4,6 +4,7 @@ import { join } from 'node:path';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { ApiError } from '../lib/errors.js';
 import { DRY_RUN_BANNER, resetDryRunBannerForTesting } from '../lib/client-factory.js';
+import { MAX_SECRET_FILE_BYTES } from '../lib/secret-file.js';
 import {
   type CliProject,
   type CliCreateProjectResponse,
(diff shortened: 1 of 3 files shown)


naufalfx805-source:

zeshi-du:
Closing this one, and I want to give you the actual reason rather than a silent close, because you did the work well.

The size cap is **declined by a decision that was recorded on #58** after this PR was opened. The relevant part:

> **No size cap on the secret-file flags.** The cap on `--code-file` exists because test code is genuinely large and unbounded; a secret is not, and adding a ceiling to already-shipped flags is a behaviour change rather than part of fixing a crash — someone with a legitimately unusual token file would start getting exit 5 on an upgrade. The guard's job here is to turn a typo into a typed error, not to police size.

The crash you were protecting against is already handled: `readSecretFileGuarded` in `src/lib/secret-file.ts` (landed in #302) turns every filesystem failure into a typed `VALIDATION_ERROR` attributed to the flag the user typed, rejects directories up front, and — the case closest to yours — rejects non-regular files via `statSync().isFile()`, so device files and FIFOs can't block a read forever.

So the hole is closed, just by a different mechanism than the one you proposed. Thank you for the report and the patch; the reasoning in it was sound, and it's on us that the decision landed on a different issue while your PR sat here.

PR #323, opened 2026-08-27
fix(auth): verify credential lock ownership

### LINKED ISSUE #77: [Hackathon] Concurrent Profile Writes Silently Drop Updates (No File Lock)

naufalfx805-source:
## Summary
Verifies credential lock ownership and locks credential profile mutations to prevent race conditions.

Closes #77.

## Changes
- Added credential lock ownership verification
- Locked credential profile mutations

## Testing
- Tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential profile updates to prevent conflicts when multiple processes modify credentials.
  * Automatically recovers from abandoned credential locks.
  * Deleting a profile now accurately reports whether a profile was removed.
  * Deleting a missing profile no longer creates an unnecessary credentials file.

### DIFF
--- a/src/lib/credentials.test.ts
+++ b/src/lib/credentials.test.ts
@@ -157,6 +157,19 @@ describe('writeProfile', () => {
     expect(file.dev).toEqual({ apiKey: 'sk-dev', apiUrl: 'https://dev' });
   });
 
+  it('reclaims stale mutation locks and removes the lock after writing', () => {
+    const lockPath = `${credentialsPath}.lock`;
+    writeFileSync(
+      lockPath,
+      `${JSON.stringify({ pid: process.pid, createdAt: Date.now() - 60_000, token: 'stale' })}\n`,
+    );
+
+    writeProfile('default', { apiKey: 'sk-new' }, { path: credentialsPath });
+
+    expect(readProfile('default', { path: credentialsPath })).toEqual({ apiKey: 'sk-new' });
+    expect(existsSync(lockPath)).toBe(false);
+  });
+
   it('does not leak the api key into the on-disk file format aside from the value itself', () => {
     writeProfile('default', { apiKey: 'sk-secret-12345' }, { path: credentialsPath });
     const onDisk = readFileSync(credentialsPath, 'utf-8');
@@ -168,6 +181,7 @@ describe('writeProfile', () => {
 describe('deleteProfile', () => {
   it('returns false when the profile is missing', () => {
     expect(deleteProfile('nope', { path: credentialsPath })).toBe(false);
+    expect(existsSync(credentialsPath)).toBe(false);
   });
 
   it('removes the named profile and leaves others intact', () => {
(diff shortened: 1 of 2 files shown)


naufalfx805-source:

naufalfx805-source:
Closing as a duplicate — this change already landed in `main` via d821dac (#272). This branch was cut from an older base, which is why it showed as conflicting. Apologies for the noise.

PR #317, opened 2026-08-20
fix(skill-nudge): log swallowed errors in debug mode

stuart902:
## What does this PR do?

Threads the existing global `--debug` flag into the skill-nudge pre-action context so the two remaining swallowed error paths in `src/lib/skill-nudge.ts` are diagnosable when debug output is requested.

Specifically:

- Profile lookup failures now emit a debug-only stderr diagnostic instead of disappearing silently.
- Managed-section target read failures (for example, an unreadable `AGENTS.md`) emit a debug-only stderr diagnostic while continuing to treat that target as absent.
- Non-debug behavior remains unchanged.
- JSON output remains unpolluted.
- Diagnostic delivery is itself best-effort, so a failing stderr/diagnostic sink cannot break the command.
- Regression tests cover both swallowed-error paths and the failure-isolation behavior.

This follows the debug-emission pattern established in #223.

## Related issue

Closes #285

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation only
- [ ] Build / CI / chore

## Checklist

- [x] PR targets the `main` branch.
- [x] Commits follow [Conventional Commits](https://www.conventionalcommits.org)
      (`feat(...)`, `fix(...)`, `docs(...)`, …).
- [x] `npm run lint` and `npm run format:check` pass.
- [x] `npm run typecheck` passes.
- [x] `npm test` passes and coverage stays at or above the 80% gate.
- [x] New behavior is covered by unit tests (mock-based; no network or
      credentials required).
- [x] No secrets, API keys, internal endpoints, or personal data are included.
- [x] User-facing changes are reflected in `README.md` / `DOCUMENTATION.md` where
      relevant.

## Notes for reviewers

The implementation is intentionally narrow and follows #223's existing pattern:
capture the swallowed error, honor the existing `--debug` flag, emit to stderr,
and preserve normal output and exit behavior.

The managed-target read failure occurs inside `isVerifySkillInstalled()`, so this
PR adds an optional best-effort diagnostic callback rather than changing the
function's return behavior. The callback is isolated with its own error guard so
diagnostic failures cannot affect command execution.

The branch changes only:

- `src/index.ts`
- `src/lib/skill-nudge.ts`
- `src/lib/skill-nudge.test.ts`

Full lint, format, typecheck, unit-test, and coverage verification is left to the
repository CI on this PR.

[...]

PR #325, opened 2026-08-27
fix: redact presigned urls in text output

### LINKED ISSUE #86: [Hackathon] Redact presigned video/analysis URLs in text output

naufalfx805-source:
## Summary

The text renderer printed `videoUrl` and `failureAnalysisUrl` verbatim — signature query
string included. Those are presigned S3 URLs, so the query *is* the credential: anyone who
can read a terminal capture, a CI log, or a pasted snippet can fetch the artifact until the
signature expires.

Closes #86.

## Changes

- Added `redactUrl()` — parses the URL and clears `username`, `password`, `search` and
  `hash`, returning the input unchanged if it does not parse.
- Applied it to `videoUrl` and `failureAnalysisUrl` in both text renderers.

`targetUrl` is deliberately **not** redacted: that is the URL the user asked to test, and
its query is meaningful to them rather than a credential.

`--output json` is untouched, so callers that need to download the artifact still get the
working URL. Only the human-readable rendering drops the signature.

## Testing

- `src/commands/test.test.ts` — 356/356 pass. Assertions cover both directions: the text
  block no longer contains `?X` / `?sig`, and the JSON payload still does.
- `tsc --noEmit` clean.

## Note

Rebased onto `main` at v0.7.0 — the previous head was cut from an older base and showed as
conflicting. The conflict was in `renderFailureText` / `renderResultText`, where `main` had
since added `appendBackendArtifactLines`; resolved by keeping that call and wrapping the
URLs.

### DIFF
--- a/src/commands/test.test.ts
+++ b/src/commands/test.test.ts
@@ -4236,6 +4236,8 @@ describe('runResult', () => {
     expect(seen[0]).toContain('/tests/test_fe/result');
     expect(got).toEqual(RESULT_FAILED);
     expect(JSON.parse(out[0]!)).toEqual(RESULT_FAILED);
+    expect(out[0]).toContain('video/run_abc.mp4?X');
+    expect(out[0]).toContain('analysis/run_abc.json?X');
   });
 
   it('text mode for a failed run highlights failureKind + failedStepIndex up top', async () => {
@@ -4265,7 +4267,14 @@ describe('runResult', () => {
     expect(block).toContain('verdict:            failed');
     expect(block).toContain('executionStatus:    completed');
     expect(block).toContain('summary:            Failed (assertion) on step 5');
-    expect(block).toContain('failureAnalysisUrl: ');
+    expect(block).toContain(
+      'videoUrl:           https://s3-presigned.example.com/video/run_abc.mp4',
+    );
+    expect(block).toContain(
+      'failureAnalysisUrl: https://s3-presigned.example.com/analysis/run_abc.json',
+    );
+    expect(block).not.toContain('video/run_abc.mp4?X');
+    expect(block).not.toContain('analysis/run_abc.json?X');
   });
 
   it('text mode for a passed run skips failureKind/failedStepIndex/failureAnalysisUrl', async () => {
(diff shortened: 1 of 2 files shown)


naufalfx805-source:

zeshi-du:
Closing this, with the reasoning and — importantly — a way forward, because the idea is not wrong, only mis-aimed.

**Why not on the success path.** This was triaged on #86 on 2026-07-03 and closed not-planned:

> in text mode those URLs are exactly what a human needs to click, and stripping the signature turns them into dead links; the short TTL is the leak mitigation by design.

A presigned S3 URL with its signature removed is a 403. So redacting on the path where we're deliberately handing someone a link to open trades a working feature for the appearance of safety. `--output json` stays verbatim for machine consumers for the same reason.

**Where it is wanted.** That same triage note left one door explicitly open:

> A `--debug`-stream-only redaction would be considered as a separate proposal.

That's the version worth building. Redaction already exists on the error/diagnostic paths (`bundle.ts` does it, landed via #60), so the pattern and the helper are there — the gap is debug output specifically. #317 is currently adding exactly that class of debug logging, so there'll be a natural surface to apply it to once it lands.

If you want to take that, open an issue describing the `--debug`-only scope and I'll assign it to you. Sorry this sat unanswered for three weeks.

PR #328, opened 2026-08-27
fix: reject malformed api keys

### LINKED ISSUE #93: [Hackathon]Malformed API key (autocorrected dash / smart quote / newline) surfaces as opaque UNAVAILABLE after wasted retries

Davidson

Add a client-side guard so a malformed API key fails fast with a clear VALIDATION_ERROR (exit 5) instead of an opaque, misleading UNAVAILABLE (exit 10) emitted only after the full retry-and-backoff cycle.

The CLI sends the key as the x-api-key HTTP header. If the key contains a character that is invalid in an HTTP header value — most commonly an autocorrected dash (en/em-dash, codepoint > 255), a smart quote, an emoji, or an embedded newline — Node's fetch throws TypeError: Cannot convert argument to a ByteString ... deep in the transport layer. That throw is classified as a network error and retried with exponential backoff, then surfaced as a generic "service temporarily unavailable" — even though retrying can never succeed and the real problem is the key.

These artifacts are exactly what you get when a key is copy-pasted from a doc, email, or chat where autocorrect silently rewrote a hyphen to an en-dash.

$ export TESTSPRITE_API_KEY="sk-user-abc–def"
$ testsprite project list --output json
{
  "error": {
    "code": "UNAVAILABLE",
    "message": "Cannot convert argument to a ByteString because the character at index 11 has a value of 8211 which is greater than 255.",
    "nextAction": "Service is temporarily unavailable. The CLI retries with exponential backoff; if this persists, report it.",
    "details": {}
  }
}
# exit 10 — and it took ~2s because of the wasted retry backoff

### Confirmations

naufalfx805-source:
## Summary
Validates API key format before use to prevent cryptic downstream errors.

Closes #93.

## Changes
- Added API key format validation

## Testing
- Tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for API keys before making requests.
  * Invalid, empty, whitespace-only, or unsupported-character keys now produce a clear validation error instead of triggering network requests or retries.
  * Normal API keys continue to work as expected.

### DIFF
--- a/src/lib/client-factory.test.ts
+++ b/src/lib/client-factory.test.ts
@@ -1,7 +1,8 @@
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import {
   DRY_RUN_API_KEY,
   DRY_RUN_BANNER,
+  assertValidApiKeyHeaderValue,
   assertValidEndpointUrl,
   emitDryRunBanner,
   makeHttpClient,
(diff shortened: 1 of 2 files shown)


naufalfx805-source:

naufalfx805-source:
Closing as a duplicate — this change already landed in `main` via b5879de (#141). This branch was cut from an older base, which is why it showed as conflicting. Apologies for the noise.

PR #336, opened 2026-08-28
fix(run): render n/a for an empty frontend stepSummary

### LINKED ISSUE #335: bug: a passing frontend run prints `steps 0/0 (passed=0, failed=0)` — the string the backend guard exists to avoid
## Summary

A frontend run can finish `status passed` while the run card prints
`steps 0/0 (passed=0, failed=0)` — the exact string this codebase already
identifies as misleading, and already suppresses, but only for backend tests.

Observed on v0.7.0, free tier, V3 routing, against a public static site.

```
runId       2bd760ae-9b86-45dc-bb9b-ceb7da6b304e
testId      a358c5bb-b94c-4be3-ad48-2aa9e0e008cd
status      passed
steps       0/0 (passed=0, failed=0)
```

Read literally that says "nothing executed, and it passed."

## The zeros disagree with the server's own step log

For that same run, `testsprite test steps` reports one recorded step:

```
INDEX  ACTION    STATUS  DESCRIPTION
  1    navigate  passed  Navigate to https://…
```

So the all-zero `stepSummary` on the run response is not a faithful picture of
what the backend recorded — a step exists in the step log but is counted
nowhere. (`test result` for the same run returns `stepSummary: null`, so the
zeroed object is specific to the run/poll response.)

## Why this is a CLI-side fix

`renderRunResponseText` in `src/commands/test.ts` already treats
`0/0 (passed=0, failed=0)` as a bad thing to print — its own comments say so:

> BE tests are atomic — no per-step breakdown exists; showing
> `0/0 (passed=0, failed=0)` reads like a no-op rather than a passing/failing
> atomic result.

and at the call site:

> Show n/a instead of confusing "0/0 (passed=0, failed=0)" which reads like a
> broken no-op.

But the guard is keyed on `isBackend`, not on the summary being empty:

```ts
if (isBackend) {
  lines.push(`steps       n/a (backend)`);
} else if (run.stepSummary) {
  lines.push(
    `steps       ${run.stepSummary.completed}/${run.stepSummary.total} (passed=${run.stepSummary.passedCount}, failed=${run.stepSummary.failedCount})`,
  );
}
```

A frontend run whose `stepSummary` comes back all zeros therefore prints the
very line the guard exists to prevent. The `null` case is already handled — the
`else if` skips it — so this is only the zeroed-object case falling through.

## Suggested fix

Key the suppression on the summary being empty rather than on the test type, so
frontend gets the same protection backend already has:

```ts
if (isBackend) {
  lines.push(`steps       n/a (backend)`);
} else if (run.stepSummary && run.stepSummary.total > 0) {
  lines.push(`steps       ${…}`);
} else if (run.stepSummary) {
  lines.push(`steps       n/a (no per-step breakdown reported)`);
}
```

That keeps every existing rendering identical and changes only the case that
currently reads as a broken no-op. Happy to open the PR with tests if this is
accepted — `/assign` me if so.

## Context, not part of the ask

The plan behind this test asked for a comparison of two numbers rendered on the
page. On the first execution of the same plan
(`testId c209acfb-937c-45a9-bf5a-68480834fa13`) the generated code was:

```python
await page.goto("VAR_{url}")
current_url = await page.evaluate("() => window.location.href")
assert current_url, 'Page should have loaded with a URL'
```

— no read of either number, verdict `passed`, with a comment saying the outcome
was "verified by the AI judge during the run". Since reruns replay the saved
script, that stored artifact would stay green regardless of the values it was
created to compare. That part is agent/platform behaviour rather than the CLI,
so I've kept it out of the fix above, but it is the reason the `0/0` line was
worth chasing: the run card is the only place a user would notice that nothing
was really checked.

naufalfx805-source:
## Summary

A frontend run whose `stepSummary` comes back all zeros printed
`steps 0/0 (passed=0, failed=0)` next to `status passed` — which reads as
"nothing executed, and it passed".

Closes #335.

## Why the existing guard missed it

`renderRunResponseText` already treats that exact string as bad output. Its own
comment says so:

> BE tests are atomic — no per-step breakdown exists; showing
> `0/0 (passed=0, failed=0)` reads like a no-op rather than a passing/failing
> atomic result.

But the guard was keyed on `isBackend`, not on the summary being empty, so a
frontend run with a zeroed summary fell straight through to the verbatim
render. The `null` case was already covered by the existing `else if
(run.stepSummary)`; only the zeroed-object case was unprotected.

## Change

One branch added. A zeroed summary is an **absent** breakdown, not a breakdown
of zero steps, so it now says that:

```
steps       n/a (no per-step breakdown reported)
```

Every other rendering is byte-identical — backend still shows `n/a (backend)`,
a real summary still shows its counts, and a `null` summary still omits the
line entirely.

## Testing

- New regression test in the existing `C2` describe block. Verified it **fails
  on `main`** with `expected '…' to contain 'steps n/a (no per-step breakdown…'`
  against the actual `status passed` / `steps 0/0 (passed=0, failed=0)` output,
  and passes with the fix.
- `src/commands/test.run.spec.ts` — 116/116 pass.
- `tsc --noEmit` clean; Prettier clean; ESLint Security unchanged at the
  baseline count for both touched files (27 before, 27 after — all pre-existing).

## How it was found

Running the CLI against a live static site on the free tier. The run came back
`passed` with `steps 0/0`, while `testsprite test steps` for the same run showed
one recorded `navigate` step — so the zeros did not match the server's own step
log either. Details and run ids are in #335.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected text-mode test results for frontend runs with no per-step breakdown.
  * These results now display “steps n/a (no per-step breakdown reported)” instead of the misleading “0/0” count.
  * Backend placeholder messaging and non-empty step summaries remain unchanged.

* **Tests**
  * Added coverage to verify the updated output and prevent regression.

### DIFF
--- a/src/commands/test.run.spec.ts
+++ b/src/commands/test.run.spec.ts
@@ -1635,6 +1635,74 @@ describe('C2 — backend run renders steps: n/a (backend) in text mode', () => {
     expect(out).not.toContain('n/a (backend)');
   });
 
+  it('FE run with an all-zero stepSummary renders n/a, not 0/0 (#335)', async () => {
+    // A passing frontend run whose summary comes back zeroed. Rendering it
+    // verbatim prints `steps 0/0 (passed=0, failed=0)` beside `status passed`,
+    // which reads as "nothing executed, and it passed" — the same no-op wording
+    // the backend branch already guards against. The zeros are not a breakdown
+    // of zero steps; they are an absent breakdown.
+    const { credentialsPath } = makeCreds();
+    const feRun: RunResponse = {
+      runId: 'run_fe_zero',
+      testId: 'test_fe_zero',
+      projectId: 'p1',
+      userId: 'u1',
+      status: 'passed',
+      source: 'cli',
+      createdAt: '2026-05-15T10:00:00.000Z',
+      startedAt: '2026-05-15T10:00:01.000Z',
+      finishedAt: '2026-05-15T10:00:30.000Z',
+      codeVersion: 'v1',
+      targetUrl: 'https://example.com',
+      createdFrom: null,
+      failedStepIndex: null,
+      failureKind: null,
+      error: null,
+      videoUrl: null,
+      stepSummary: { total: 0, completed: 0, passedCount: 0, failedCount: 0 },
+    };
+    const handler = (url: string) => {
+      if (url.includes('/tests/test_fe_zero/runs')) {
+        return {
+          body: {
+            runId: 'run_fe_zero',
+            status: 'queued',
+            enqueuedAt: '2026-05-15T10:00:00.000Z',
+            codeVersion: 'v1',
+            targetUrl: 'https://example.com',
+          },
+        };
+      }
+      if (url.includes('/runs/run_fe_zero')) return { body: feRun };
+      return { status: 404, body: {} };
+    };
+    const stdoutLines: string[] = [];
+    await runTestRun(
+      {
+        profile: 'default',
+        output: 'text',
+        debug: false,
+        dryRun: false,
+        testId: 'test_fe_zero',
+        wait: true,
+        timeoutSeconds: 60,
+      },
+      {
+        credentialsPath,
+        fetchImpl: makeFetch(handler),
+        stdout: line => stdoutLines.push(line),
+        stderr: () => {},
+        sleep: instantSleep,
+      },
+    );
+    const out = stdoutLines.join('\n');
+    expect(out).toContain('steps       n/a (no per-step breakdown reported)');
+    // The string the backend branch exists to avoid must not appear here either.
+    expect(out).not.toContain('0/0');
+    // It is a frontend run, so it must not claim to be a backend one.
+    expect(out).not.toContain('n/a (backend)');
+  });
+
   it('BE run terminal on first poll (opts.type=backend) renders steps: n/a via type hint (Fix 2)', async () => {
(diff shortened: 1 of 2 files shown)


zeshi-du (approved):
Verified against current `main` (v0.11.0): merges cleanly, full CI green.

The diagnosis in #335 is right — the guard was keyed on `isBackend` rather than on the summary being empty, so the one case it exists to prevent (`0/0 (passed=0, failed=0)` next to a `passed` status) fell straight through on the frontend path. Splitting it into `stepSummary && total > 0` / `stepSummary` is the minimal correct fix, and the wording you chose ("a zeroed summary is an absent breakdown, not a breakdown of zero steps") is exactly the distinction.

One thing for the record, because merging this closes #335 and the server-side half should not disappear with it. The zeros are not only a rendering problem: for that run, `test steps` returns a recorded `navigate` row while the run response reports `total: 0`. **We've filed the wire inconsistency internally as its own tracked bug** — the run/poll response disagreeing with the step log for the same `runId`. This PR is the right fix for the CLI's half; it is not a fix for the backend's.

Worth knowing as a consequence: after this merges, a run with rows that weren't counted and a run with genuinely no rows both render as the same `n/a` line, so that distinction now has to be made from the API rather than from CLI output. That's an acceptable trade — printing "nothing executed, and it passed" is worse — but it's a real one.

Sorry this sat for over two weeks. Good report, good fix.

PR #291, opened 2026-07-26
feat(http): validate the /me and usage response shapes at runtime

Davidson3556:
## What does this PR do?

Extends the response-validation layer established by #266 to the **account surfaces** — the `GET /me` reads and the usage projection. This is the first of the incremental groups #277 asks for (reads / writes / account surfaces); the test and project shapes are left for follow-ups.

> Part of #277. Tracked by #292.

All three `/me` callers go through the generic `client.get`, so all three still blind-cast the body:

| caller | shape | what a drifted body did before |
|---|---|---|
| `auth status` / `whoami` | `MeResponse` | `m.scopes.join(', ')` and `m.scopes.includes(...)` are unguarded → raw `TypeError` (exit 1) in text mode, and **exit 0 with a partial identity** under `--output json` |
| `usage` / `credits` | `UsageResponse` | a string `credits` reached the `Math.floor(credits / creditsPerRun)` pre-flight arithmetic |
| `doctor` | `MeIdentity` | fully-optional projection, but its local interface had already drifted from the stubbed schema (`v3Enabled` existed on the command side only) |

Changes:

- New `ME_RESPONSE_SCHEMA` and `USAGE_RESPONSE_SCHEMA` in `src/lib/response-schemas.ts`.
- Wires the previously-unwired `ME_IDENTITY_SCHEMA` into `doctor`, and adds the missing `v3Enabled` to it.
- `doctor`'s `MeIdentity` is now a type alias of the schema's wire type, so interface and schema cannot drift apart again.
- Schemas are passed at the three `client.get('/me', { schema })` call sites. `init` inherits validation through `runWhoami`, where the existing try/catch already degrades to a placeholder identity — so a drifted `/me` never fails the whole `init`.

### Policy compliance (#266, unchanged)

- Every object is `looseObject`: additive server fields pass **and** are preserved, so `--output json` stays byte-faithful. Covered by a test that sends an unknown `orgName` through `whoami --output json`.
- `env` is validated as an open string via `openWireLiteral`, so a new deployment tier (`sandbox`, …) cannot hard-fail the CLI.
- Fixture-evidence discipline for required vs optional:
  - **Required** — `userId`, `keyId`, `scopes`, `env`. Every `/me` fixture in the suite supplies all four (`auth.test.ts`, `init.test.ts`, `usage.test.ts`, `cli.subprocess.test.ts`, `test/mock-backend/fixtures.ts`, `lib/dry-run/samples.ts`), and `scopes` is exactly the field whose absence crashes today.
[...]

Issue #277, opened 2026-07-23
Extend runtime response validation to the remaining typed response shapes


zeshi-du:
Follow-up to #266 (merged), which established the validation policy (`looseObject` + `openWireLiteral` + evidence-based `nullish`/`optional` — see the merge comment there) and wired it into the 7 run-path shapes.

Remaining work: bring the other typed response shapes onto the same schema layer incrementally — `CliTest`, `CliTestCode`, `CliCreateTestResponse`, `CliProject`, `MeResponse` (schema already stubbed as `ME_IDENTITY_SCHEMA`, currently unwired), `UsageResponse`, `CliFailureSummary`, and the rest of the `as T` casts in `src/lib/http.ts` helpers.

Guidelines:
- Follow #266's policy exactly; no fail-closed on fields the server may legitimately add or omit.
- One PR per coherent group of shapes is fine (reads vs writes vs account surfaces) — no need for a single mega-PR.
- Each schema needs the same fixture-evidence discipline #266 used (cite the test fixture that proves a field is optional vs nullable).

Good structured task for anyone who wants it — comment /assign to claim.

Davidson3556:
Started on this. PR #291 covers the **account surfaces** group: `ME_RESPONSE_SCHEMA` + `USAGE_RESPONSE_SCHEMA`, and it wires the stubbed `ME_IDENTITY_SCHEMA` into `doctor` (whose local `MeIdentity` had already drifted — it carried `v3Enabled`, the schema didn't).

The one that actually bites: `whoami` calls `m.scopes.join(', ')` unguarded, so a `/me` body without `scopes` throws a raw TypeError in text mode and, under `--output json`, exits 0 with a partial identity. Before/after is in the PR.

Next up, unless you'd rather have them ordered differently: the read shapes (`CliTest`, `CliTestCode`, `CliProject`), then `CliCreateTestResponse` + `CliFailureSummary` and the remaining `as T` casts in the `http.ts` helpers.

zeshi-du:
Umbrella status.

Tracked internally as **DEV-879** — runtime response validation is missing for `CliTest` / `CliTestCode` / `CliCreateTestResponse` / `CliProject` / `CliFailureSummary`, which trust the wire unchecked. This is confirmed and wanted; it isn't a speculative hardening request.

Two notes for the people working under this umbrella:

[...]

PR #340, opened 2026-09-06
fix: validate project and test-code responses before use

jerrybate-sketch:
## What does this PR do?

Malformed project and test-code responses can currently appear successful, crash the command, or replace an existing output file. This change checks those responses before use and returns the normal `INTERNAL` error with the mismatched field paths.

Extra server fields and future enum values still pass through. Missing optional project fields remain missing, and draft code with a null body follows the existing no-code path.

## Related issue

Part of #277, which is assigned to me and permits one PR per coherent group. This covers project list/get and standalone test-code reads, including the version read before `code put`. Other response groups remain open; account validation in #291 is outside this change.

I have left the umbrella issue open. The issue gate asks for a closing reference; please let me know whether you prefer a separate tracking issue for this group or an exception for the umbrella link.

## Type of change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation only
- [ ] Build / CI / chore

## Checklist

- [x] PR targets `main`.
- [x] Conventional Commit.
- [x] Lint and formatting pass.
- [x] Typecheck passes.
- [x] Full unit suite passes; coverage exceeds the 80% gate.
- [x] New behavior has mocked regression tests.
- [x] No secrets, private endpoints or personal data are included in the patch.
- [x] No new command, flag or setup step needs documentation.

## Notes for reviewers

The new behavior tests reproduce failures on the upstream code. Existing fixtures establish the accepted optional and nullable fields; comments point to that evidence. No runtime dependency was added.

Local checks ran on Windows with Node 24.19.0. The suite needed normal temporary-file access and an empty `NO_COLOR` setting for its existing terminal tests. The remote Linux/Node 20 and 22 checks remain for CI.

## Summary by CodeRabbit

- **Bug Fixes**
  - Added stronger validation for project and test-code responses, preventing malformed data from being displayed or written.
  - Improved handling of legacy tests where source code or framework information may be unavailable.
  - Preserved additional project and test metadata, including newer field values, without inventing missing optional information.
  - Validated every page during automatic pagination to ensure consistent results across complete project lists.
  - Improved compatibility with newly introduced language and framework values returned by the service.

### DIFF
--- a/src/commands/project.test.ts
+++ b/src/commands/project.test.ts
@@ -631,6 +631,79 @@ describe('createProjectCommand --page-size option parser', () => {
   });
 });
 
+describe('project read response validation', () => {
+  it.each([
+    ['missing name', { ...PROJECT_FIXTURE, name: undefined }],
+    ['invalid attribute list', { ...PROJECT_FIXTURE, testIdAttributes: 'data-testid' }],
+    ['invalid nullable URL', { ...PROJECT_FIXTURE, targetUrl: false }],
+  ])('rejects %s before printing a successful result', async (_label, body) => {
+    const out: string[] = [];
+    await expect(
+      runGet(
+        { profile: 'default', output: 'json', debug: false, projectId: PROJECT_FIXTURE.id },
+        { ...makeCreds(), fetchImpl: makeFetch(() => ({ body })), stdout: line => out.push(line) },
+      ),
+    ).rejects.toMatchObject({ code: 'INTERNAL', exitCode: 1 });
+    expect(out).toEqual([]);
+  });
+
+  it.each([false, true])('validates list rows before output (single page: %s)', async single => {
+    const out: string[] = [];
+    let calls = 0;
+    const fetchImpl = makeFetch(() => {
+      calls += 1;
+      return {
+        body:
+          !single && calls === 1
+            ? { items: [PROJECT_FIXTURE], nextToken: 'second-page' }
+            : { items: [{ ...PROJECT_FIXTURE, id: null }], nextToken: null },
+      };
+    });
+    await expect(
+      runList(
+        { profile: 'default', output: 'text', debug: false, ...(single ? { pageSize: 1 } : {}) },
+        { ...makeCreds(), fetchImpl, stdout: line => out.push(line) },
+      ),
+    ).rejects.toMatchObject({ code: 'INTERNAL', exitCode: 1 });
+    expect(calls).toBe(single ? 1 : 2);
+    expect(out).toEqual([]);
+  });
+
+  it('rejects a malformed cursor instead of returning it as a usable next page', async () => {
+    const out: string[] = [];
+    await expect(
+      runList(
+        { profile: 'default', output: 'json', debug: false, pageSize: 1 },
+        {
+          ...makeCreds(),
+          fetchImpl: makeFetch(() => ({ body: { items: [PROJECT_FIXTURE], nextToken: 17 } })),
+          stdout: line => out.push(line),
+        },
+      ),
+    ).rejects.toMatchObject({ code: 'INTERNAL', exitCode: 1 });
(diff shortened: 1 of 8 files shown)


jerrybate-sketch:
The type fix is in e5bab11. Code reads now expose the strings the server can actually return, with compiler checks in the existing compatibility test.

P4 remains strict because it checks the documented fixtures; separate tests cover draft, legacy and future responses. Loosening P4 would remove that independent check. TypeScript and all 426 affected tests pass.

We planned and researched with AI.

zeshi-du:
You asked a direct question nine days ago — whether to open a separate tracking issue for this group or get an exception for the umbrella link — and got no answer. Apologies. You're right about the mechanics: the gate wants `Closes #NNN` against an issue assigned to you, and "Part of #277" doesn't satisfy it, so the red check is our configuration being blunt rather than anything wrong with your PR. I'm settling that this week and will come back with a concrete instruction, not another round of silence.

Meanwhile, two things you can act on now.

**Your CI had never run.** Fork PRs here wait at `action_required` for a maintainer to release them, and nobody did — for nine days. I've released it. Typecheck, unit tests (Node 20/22 + Windows), build, coverage, E2E, lint/format and the secret scans are all green.

**One real failure:** `ESLint Security (changed files)` flags four `security/detect-non-literal-fs-filename` findings in `src/commands/test.test.ts` — lines 1579, 1589, 1590 and 1633 (`writeFileSync`, `readFileSync`, `readdirSync`, `writeFileSync`). The repo's convention for deliberate tmpdir-based fs calls in tests is an `// eslint-disable-next-line security/detect-non-literal-fs-filename` carrying a one-line justification; there are existing examples throughout the test suite.

**And a rebase:** `main` moved to v0.11.0 while this sat, and `src/lib/response-schemas.ts` now conflicts.

(The `Dependency Review` red is broken repo-wide for reasons unrelated to any PR — a setting on our side I'm getting fixed. Ignore it.)

zeshi-du:
Follow-up with the concrete answer I promised — and it's better than what I implied earlier. **Don't restructure anything.**

The gate itself has been fixed. It now accepts non-closing references (`Part of #N`, `Refs #N`, `References #N`, `Related to #N`) in addition to closing keywords, precisely because an umbrella issue that invites one PR per coherent group — as #277 does in its own body — must not be closed by any single child PR, so there was no correct keyword you could have written. Your PR body was right and the gate was wrong.

The fix is merged internally and ships to this repo with the next release, which is in flight now. Once it lands, **`Part of #277` as you already wrote it will satisfy the gate** — you don't need a new issue and you don't need to change the link. The check will clear on your next push, which you need anyway for:

- the rebase onto current `main` (`src/lib/response-schemas.ts` conflicts), and
- the four `security/detect-non-literal-fs-filename` findings in `src/commands/test.test.ts`.

To be explicit about what this means for you: nothing you did here was wrong, you waited nine days on a contradiction in our own tooling, and the gate has been changed rather than you being asked to work around it. Thanks for flagging it in the PR body instead of quietly restructuring — that's what got it fixed.

zeshi-du (changes requested):
Hi @jerrybate-sketch — thanks for the response validation on project/test-code reads, that's a good defensive addition. Before merging:

- The `needs-issue` label and `gate` failure are stale: the issue-reference check now accepts "Part of #N" wording like yours, it just hasn't re-run since that change shipped. A small edit to the PR description (any wording tweak) will retrigger it — no separate tracking issue needed.
- The Security workflow's ESLint check (`security/detect-non-literal-fs-filename`) flags four new lines in `src/commands/test.test.ts` (~1579, 1589, 1590, 1633) for non-literal fs access. Please add the usual narrow suppression comments for those fixture writes, same as elsewhere in the suite.
- Please rebase `src/lib/response-schemas.ts`, which now conflicts with `main`.

Should be quick to land once those are in.

PR #342, opened 2026-09-12
docs: update manual authentication commands

xujiantop-crypto:
## What does this PR do?

Updates the manual authentication walkthrough to use the current, visible CLI commands:

- `testsprite setup --no-agent` configures credentials without installing an agent skill.
- `testsprite auth status` verifies the active identity and profile.

This keeps the manual, step-by-step flow intact while avoiding hidden deprecated aliases and their deprecation notices.

## Related issue

N/A - documentation-only correction; the contribution guide does not require an issue for doc fixes.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Documentation only
- [ ] Build / CI / chore

## Checklist

- [x] Commits follow Conventional Commits.
- [x] `npm test` is not required because runtime behavior is unchanged.
- [x] New behavior tests are not applicable because runtime behavior is unchanged.
- [x] User-facing changes are reflected in `DOCUMENTATION.md`.

## Notes for reviewers

`npm run lint` completes with zero errors; its existing warnings are unrelated to this documentation-only change. `npm run format:check` and `npm run typecheck` pass.

## Summary by CodeRabbit

- **Documentation**
  - Updated authentication setup instructions to use the current setup commands.
  - Added the `--no-agent` option to environment-based setup guidance.
  - Renamed the authentication verification command to `testsprite auth status`.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -64,13 +64,13 @@ The CLI uses API keys. Create one from your [TestSprite dashboard](https://www.t
 
 ```bash
 # Interactive — prompts for your API key (input is masked); endpoint defaults to prod
-testsprite auth configure
+testsprite setup --no-agent
 
 # Non-interactive — reads TESTSPRITE_API_KEY from the environment (CI / scripts)
-TESTSPRITE_API_KEY=sk-... testsprite setup --from-env
+TESTSPRITE_API_KEY=sk-... testsprite setup --from-env --no-agent
 
 # Verify
-testsprite auth whoami
+testsprite auth status
 ```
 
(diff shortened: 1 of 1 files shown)


zeshi-du (approved):
Correct, and verified against current `main` (v0.11.0).

`auth.ts` marks both of the commands this walkthrough was teaching as hidden, deprecated aliases: `auth configure` for `setup`, and `auth whoami` for `auth status`. So the manual-authentication section was sending people to two commands that don't appear in `--help` and print a deprecation notice when run — while the paragraph a few lines further down already refers to `auth status`. Adding `--no-agent` to the `--from-env` line is right too: it matches the note directly beneath it about using `--no-agent` when only session authentication is needed.

Checked locally rather than in CI: the resulting file passes `prettier --check`, and nothing outside `DOCUMENTATION.md` is touched.

Thanks — small, exact, and it fixes a page that new users hit first.

